PR #2, opened 2020-12-02
Add validations

IreneMLC:
- Add validations with some documentation.
- Add github actions, PR template and .gitignore.

After this PR is merged I'll activate the actions so to verify that they run in the next PR (which will contain the first entity definitions)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,19 @@
+# Default ignored files
+/shelf/
+/.idea/workspace.xml
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
+# Editor-based HTTP Client requests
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -0,0 +1,19 @@
+# Default ignored files
+/shelf/
+/.idea/workspace.xml
```
srvaroa, line 3: nit: this one is unnecessary with the one at L19

srvaroa (approved):

srvaroa (approved):

.gitignore
IreneMLC: Since this is merged I'll remove it in the next PR, thanks!

PR #4, opened 2020-12-02
Update definition schema to allow non-synthesizable entities

IreneMLC:
### Relevant information

Updated the entity definition schema so that we can start composite metrics for entities that will not be synthesized from this repository. 

### Checklist

* [x] This is not my first contribution, or I've reached out to the team by opening an issue before
 opening this PR. 
 explanation above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,13 @@ so your PR will get merged faster, and you can start enjoying your shiny new ent
 We use the `domain`, `type` and `identifier` to assign each entity a Global Unique Identifier (GUID).
 - The `domain` must be a value matching `/[A-Z][A-Z0-9_]{2,7}/`. This field is mostly relevant internally for NR. 
 Use EXT by default, although we may advise to use a different value in some cases.                
-- The `type` must be a value matching `/[A-Z][A-Z0-9_]{2,19}/`. This field is meant to identify the type of entity. 
+- The `type` must be a value matching `/[A-Z][A-Z0-9_]{2,11}/`. This field is meant to identify the type of entity. 
 Some examples are APPLICATION, HOST or CONTAINER.  
 - The `identifier` must be assigned a parameter that is unique within the domain and type (e.g
 . cluster ID, host ID, etc). Keep in mind that the value has the following restrictions and that the
  entity will not be synthesized if the value extracted from the metrics is considered invalid:
-  - `/[\x20-\x7E]{1,28}/`.
-  - 1 to 28 standard ascii characters, excluding control chars (codes: 32-126).
+  - `/[\x20-\x7E]{1,36}/`.
+  - 1 to 36 standard ascii characters, excluding control chars (codes: 32-126).
   - If you suspect that your identifiers may not fulfil our length requirements, set the optional `encodeIdentifierInGUID` field to true. 
 - The definition needs to provide enough information to differentiate this entity
  from others. It cannot be a subset nor a superset of any existing definition. If the names of
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Some examples are APPLICATION, HOST or CONTAINER.
  entity will not be synthesized if the value extracted from the metrics is considered invalid:
   - `/[\x20-\x7E]{1,28}/`.
   - 1 to 28 standard ascii characters, excluding control chars (codes: 32-126).
+  - If you suspect that your identifiers may not fulfil our length requirements, set the optional `encodeIdentifierInGUID` field to true. 
 - The definition needs to provide enough information to differentiate this entity
  from others. It cannot be a subset nor a superset of any existing definition. If the names of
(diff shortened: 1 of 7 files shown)


gvassallo (approved):

validator/schemas/entity-schema-v1.json
```diff
@@ -62,13 +62,30 @@
...
+      "required": [
+        "synthesize"
+      ]
+    }
   ],
   "properties": {
+    "synthesize": {
+      "$id": "#/properties/synthesize",
+      "type": "boolean",
+      "title": "Whether the entity should be synthesized",
```
srvaroa, line 84: Maybe "the entity type" is more accurate. 

Also, if I read this correctly a non-synth'd entity would  not have the domain/type fields, which however seem relevant anyway?

```
synthesize: false
```


I'm wondering if we should do something like this instead.  Move the synthesis config inside a `synthesis` node:

```
domain: INFRA
type: ETCD_CLUSTER

synthesis:
  identifier: clusterName  
  name: clusterName
  conditions:
    - attribute: metricName
      prefix: etcd_
  tags:
    - label.topology.kubernetes.io/region
    - ...
```

This would let us define non-synthesisable fields as:

```
domain: INFRA
type: ETCD_CLUSTER
```

(e.g. absence of a synthesis node)

What do you think?

srvaroa (commented):
Thanks @IreneMLC ! Some questions.

Also, we should update docs once we settle on the final solution.

validator/schemas/entity-schema-v1.json
IreneMLC: The domain and type are also on the folder name, so I was counting on that (I thought we'd ignore the definition file), but what you propose seems like a really clean solution so I'm happy to go with it.

cianBuckley (approved):
I like the suggestion of non synthesis section, apart from that 👍

IreneMLC:
I added the proposed changes, wdyt?

README.md
```diff
@@ -20,7 +20,7 @@ so your PR will get merged faster, and you can start enjoying your shiny new ent
 We use the `domain`, `type` and `identifier` to assign each entity a Global Unique Identifier (GUID).
 - The `domain` must be a value matching `/[A-Z][A-Z0-9_]{2,7}/`. This field is mostly relevant internally for NR. 
 Use EXT by default, although we may advise to use a different value in some cases.                
-- The `type` must be a value matching `/[A-Z][A-Z0-9_]{2,11}/`. This field is meant to identify the type of entity. 
+- The `type` must be a value matching `/[A-Z][A-Z0-9_]{2,19}/`. This field is meant to identify the type of entity.
```
IreneMLC, line 23: `KUBERNETES_APISERVER` was too long for this requirement, so I increased it. This will leave us less space for the identifier in the GUID, maybe that's why it was encoded @srvaroa?

README.md
```diff
@@ -20,12 +20,12 @@ so your PR will get merged faster, and you can start enjoying your shiny new ent
...
 - The `domain` must be a value matching `/[A-Z][A-Z0-9_]{2,7}/`. This field is mostly relevant internally for NR. 
 Use EXT by default, although we may advise to use a different value in some cases.                
-- The `type` must be a value matching `/[A-Z][A-Z0-9_]{2,11}/`. This field is meant to identify the type of entity. 
+- The `type` must be a value matching `/[A-Z][A-Z0-9_]{2,19}/`. This field is meant to identify the type of entity. 
 Some examples are APPLICATION, HOST or CONTAINER.  
 - The `identifier` must be assigned a parameter that is unique within the domain and type (e.g
 . cluster ID, host ID, etc). Keep in mind that the value has the following restrictions and that the
  entity will not be synthesized if the value extracted from the metrics is considered invalid:
-  - `/[\x20-\x7E]{1,36}/`.
+  - `/[\x20-\x7E]{1,28}/`.
```
IreneMLC, line 28: Reduced this to account for an increased entity type size.

cianBuckley (approved):

README.md
srvaroa: We should do the opposite actually, the sizes of each field are defined in the entity guid spec so we should instead make KUBERNETES_APISERVER shorter (my bad!)

srvaroa (changes requested):
LGTM just -1'ing for the resize of the type field.

README.md
srvaroa: Btw we should copy (move?) the entity guid spec here.

IreneMLC: Oh, good, then I'll change that. I think that we discussed adding the whole spec but it had very internal information so we added the relevant parts of the spec to the docs, I can do another check and see if anything relevant is missing (or moving it to another file if it's clearer).

srvaroa: Yeah you're right, looking in the text it really contains everything that's needed

srvaroa (approved):

srvaroa (approved):

PR #5, opened 2020-12-03
Added definition file for Prometheus Redis

kidk:
### Relevant information

I have customer using Prometheus heavily for Redis instances in their Kubernetes cluster. As far as I see we don't have entity synthesis for this yet. 

### Checklist

* [x] This is not my first contribution, or I've reached out to the team by opening an issue before
 opening this PR. 
 explanation above.


definitions/infra-prometheus-redis/definition.yml
```diff
@@ -0,0 +1,22 @@
+domain: infra
```
srvaroa, line 1: domain and type are meant to be uppercase (we should document this better!)

srvaroa, line 1:
```suggestion
domain: INFRA
```

definitions/infra-prometheus-redis/definition.yml
```diff
@@ -0,0 +1,22 @@
...
+type: prometheus-redis
```
srvaroa, line 2: We probably want REDIS here (e.g. this allows us to start synthesizing REDIS's from other telemetry sources, but keep the same type of entities)

definitions/infra-prometheus-redis/definition.yml
kidk: done

definitions/infra-prometheus-redis/definition.yml
kidk: That makes sense. I changed it to REDIS, let me know if it needs to be something else.

definitions/infra-prometheus-redis/golden_metrics.yml
```diff
@@ -0,0 +1,17 @@
+redisConnectionsReceivedPerSecond:
+  title: Number of connections received per second
+  unit: OPERATIONS_PER_SECOND
+  query:
+    select: avg(redis.instance.net.connectionsReceivedPerSecond)
+
+redisCommandsProcessedPerSecond:
+  title: Number of commands processed
+  unit: OPERATIONS_PER_SECOND
```
IreneMLC, line 9: The golden metrics don't need the unit field, so these can be removed

kidk: done

definitions/infra-prometheus-redis/definition.yml
```diff
@@ -0,0 +1,22 @@
...
+
+tags:
+  - redis.aofLastBgrewriteStatus
+  - redis.aofLastWriteStatus
+  - redis.port
+  - redis.rdbLastBgsaveStatus
+  - redis.version
+
+goldenTags:
+  - redis.clusterRole
```
IreneMLC, line 18: Final comment from me I think, the `redis.clusterRole` also needs to be on the `tags` section. We should update this on the docs.

kidk: Yeah I was a bit confused by the goldenTags section. Some more details in the docs would help.

IreneMLC: We'll use this experience to improve things. Thanks for the patience!

IreneMLC: Basically, the `tags` section includes all the tags that should be associated with the entity while the `goldenTags` are just a subset with the tags users consider most important.

IreneMLC (approved):

definitions/infra-prometheus-redis/dashboard.json
```diff
@@ -0,0 +1,130 @@
...
+    "grid_column_count": 12,
+    "filter": null,
+    "owner": {
+        "email": "svandamme@newrelic.com",
+        "id": 2359884
+    },
+    "widgets": [
+        {
+            "title": "Number of connections received per second",
+            "nrql": "FROM Metric SELECT average(redis.instance.net.connectionsReceivedPerSecond) SINCE 1 day ago TIMESERIES COMPARE WITH 7 days ago",
```
cianBuckley, line 13: I suspect that this somehow should isolate to the individual guid we'll be synthesising. Could that be something we patch post registration? Or could we make use of a placeholder? eg `where guid = ${synthesised_guid}` cc @srvaroa @IreneMLC

definitions/infra-prometheus-redis/dashboard.json
```diff
@@ -0,0 +1,130 @@
+{
+    "title": "Redis status",
+    "icon": "line-chart",
+    "grid_column_count": 12,
+    "filter": null,
+    "owner": {
```
cianBuckley, line 6: We might want to generate this section post registration (otherwise this email address will pop up over all accounts :D)

definitions/infra-prometheus-redis/definition.yml
```diff
@@ -0,0 +1,20 @@
+domain: INFRA
```
cianBuckley, line 1: Just to note: Registering under this domain implies that any 1st party redis entity types generated will have to abide by same domainID generation.

cianBuckley (commented):
Exciting! Looking forward to seeing how this progresses, Thanks @kidk !

definitions/infra-prometheus-redis/dashboard.json
IreneMLC: The placeholder seems like a good idea to me, since I guess that we'll have to replace the GUID when the template is requested from us.

definitions/infra-prometheus-redis/golden_metrics.yml
```diff
@@ -0,0 +1,14 @@
+redisConnectionsReceivedPerSecond:
+  title: Number of connections received per second
+  query:
+    select: average(redis.instance.net.connectionsReceivedPerSecond)
```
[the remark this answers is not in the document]
kidk: I will double check, but I think you might be right. I based this on a customer account, of which they said it used prometheus. Shame on me for not double checking.

definitions/infra-prometheus-redis/definition.yml
```diff
@@ -0,0 +1,16 @@
...
+type: REDIS
+identifier: targetName
```
kidk, line 3: I'm not too sure about this. Any ideas? I don't get a lot of dimensions when using the POMI integration.

definitions/infra-prometheus-redis/definition.yml
naxhh: Unrelated to the pr but maybe we need to move golden tags outside of the syntethization.. but more files is more burden too :(

definitions/infra-prometheus-redis/definition.yml
```diff
@@ -0,0 +1,16 @@
...
+  - attribute: metricName
+    prefix: redis_
+
+tags:
+  - clusterName
+  - targetName
+
+compositeMetrics:
+  goldenMetrics: ./golden_metrics.yml
+  summaryMetrics: ./summary_metrics.yml
```
IreneMLC, line 16:
```suggestion
synthesis:
  identifier: targetName
  name: targetName

  conditions:
    - attribute: metricName
      prefix: redis_

  tags:
    - clusterName
    - targetName

compositeMetrics:
  goldenMetrics: ./golden_metrics.yml
  summaryMetrics: ./summary_metrics.yml
```

definitions/infra-prometheus-redis/definition.yml
cianBuckley: Mmmm it's a good question.. We're not redis experts so it will be hard for someone on our side to say anything definitive here :( 

Could you perhaps clarify with the customer you're doing the definition for? Or somehow gather more data to verify?

We currently do not have any test process or pattern of support of any shape for this - I'm totally open to suggestions on how we can improve this!

definitions/infra-prometheus-redis/dashboard.json
cianBuckley: Update on this: We've decided to add this to a post process so the template definition here will remain as is. Thanks!

definitions/infra-prometheus-redis/dashboard.json
cianBuckley: This entire section will be added internally after commit - could you remove the owner section @kidk ?

definitions/infra-prometheus-redis/definition.yml
cianBuckley: @srvaroa have we decided that this should fall under EXT?

srvaroa: Yes I think we agreed on that with the infra team. 

@kidk could you update the domain to EXT, and  change the directory name to "ext-redis"? (the directory is formed by concatenating domain-type.

Thanks!

srvaroa (changes requested):
LGTM, just requesting changes for the domain.

definitions/infra-prometheus-redis/definition.yml
kidk: I'm getting this data through POMI which only gives me the cluster_name. The redis exporter itself also doesn't expose a hostname or a different name. I'm conflicted on this. 

In this case I feel the customer should add a custom label to it's data to get the right name, but targetName is definitely a good default. Is there a way to say, take this first, then this. 

In this case that could be: custom label (hostname), and as fallback targetName.

srvaroa (approved):

srvaroa:
Thanks @kidk ! I think this one is ready to merge (I'll wait for a bit in case @cianBuckley or @jfjoly have objections)

cianBuckley (approved):

PR #12, opened 2020-12-11
Migrate new entity definitions

IreneMLC:
### Relevant information

Migrate entity definitions.


definitions/infra-na/definition.yml
```diff
@@ -0,0 +1,2 @@
+domain: INFRA
+type: NA
```
IreneMLC, line 2: Always ignoring the spec 😬

srvaroa: I would omit this type as it's in practice an alias to other real types (that have their own definitions), NA exists just for historical reasons but we're not really going to attach golden metrics or similar stuff to it.

IreneMLC: Good, I'll add a rule to the migration to ignore it then, thanks! (closing this PR since there are other issues btw)

PR #14, opened 2020-12-14
thezackm/unix entity definition

thezackm:
### Relevant information

Adding entity for [Unix On-Host Integration](https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/unix-monitoring-integration)

### Checklist

* [X] This is not my first contribution, or I've reached out to the team by opening an issue before
 opening this PR. 
 explanation above.


definitions/infra-unix_host/definition.yml
```diff
@@ -0,0 +1,8 @@
+domain: INFRA
+type: UNIX_HOST
+name: hostname
+identifier: hostname
+
+compositeMetrics:
+  goldenMetrics: ./golden_metrics.yml
+  summaryMetrics: ./summary_metrics.yml
```
IreneMLC, line 8: Sorry for the changes, since the schema is still evolving, but the PR is failing because:

- These fields need to be under `synthesis` as in [this example](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/docs/example-entity-definition.yml).
- The golden and summary metrics are now arrays.

IreneMLC, line 8:
```suggestion
synthesis:
  name: hostname
  identifier: hostname

  compositeMetrics:
    goldenMetrics: 
      - ./golden_metrics.yml
    summaryMetrics: 
      - ./summary_metrics.yml
```

thezackm: understood; updates applied, thanks!

thezackm: @IreneMLC - still failing; no rush, just FYI

IreneMLC: Thanks for adding the changes! the validator seems to be complaining about the yaml format.

IreneMLC: The issue is comming from the field in backticks so you can surround it with single quotes and it should be good to go:
'\`unixMonitor:Disk\`'

thezackm: checks are passing, thanks for all of the help!

IreneMLC: Thanks for adding the changes!

definitions/infra-unix_host/definition.yml
```diff
@@ -0,0 +1,11 @@
+domain: INFRA
+type: UNIX_HOST
+synthesis:
+  name: hostname
+  identifier: hostname
```
IreneMLC, line 5: This attribute name might be a bit generic, I think it might make sense if it had some condition but wdyt @cianBuckley @srvaroa ?

definitions/infra-unix_host/definition.yml
```diff
@@ -4,8 +4,8 @@ synthesis:
   name: hostname
   identifier: hostname
 
-  compositeMetrics:
-    goldenMetrics: 
-      - ./golden_metrics.yml
-    summaryMetrics: 
-      - ./summary_metrics.yml
+compositeMetrics:
```
IreneMLC, line 7: To account for a schema change

definitions/infra-unix_host/definition.yml
thezackm: in the sample data we have for the Unix OHI; `hostname` is the most unique attribute available unfortunately.

IreneMLC: No problem, is there some condition that can be added to differentiate from another entity that might use that attribute though? for example, for redis telemetry they can use something like 'redis_' prefix.

thezackm: do you mean can we manually add a prefix to the definition to make it unique; or are you asking if there's another attribute in the payload that we can concatenate to make it unique?

the first option is open; I don't think it matters much what this looks like in our back end if it doesn't change the way our users see the data. `unix_host` vs. `hostname` isn't a huge difference IMO. 

the second option, I don't believe is viable without a change to the integration to update the way it reports data and manually create a unique identifier there.

thezackm:
@IreneMLC @jfjoly - question came up that I'm unsure how to address. Hoping you can help: 
The Unix integration covers [multiple OS types](https://github.com/newrelic/newrelic-unix-monitor/tree/master/config), and it doesn't make a lot of sense to call them all `UNIX_HOST`. i.e.; if I have both AIX and MacOS in my integration, I want to visually and logically keep those separate in NROne. 

Is it possible right now to create these definitions and limit them based on part of the event attributes? 

Something like this perhaps? 
```
If (SELECT osName FROM `unix:System` = 'aix'), Then apply this definition...
```

Ultimately I think the confusion is how granular can we get on parsing/faceting unique entities out of a comment event type.

jfjoly:
@thezackm the osName can be a golden tags that will be shown in the UI. The user can create a filter to see only the entities of this type and with the tag osName='AIX'.
If we want different entity types, then we should create multiple definitions.

definitions/infra-unix_host/definition.yml
IreneMLC: Oh, yes I meant the second one, using another payload attribute to ensure that we're able to differentiate. Thanks for confirming it.

IreneMLC:
Multiple definitions could work well if we had a metric attribute that could be used like that `unix:System`.

definitions/infra-unix_host/definition.yml
cianBuckley: I don't think our team have any definitive answer on this - if you feel hostname is sufficient to be unique across the domain, entity type and account, then that's fantastic

cianBuckley (approved):

definitions/infra-unix_host/definition.yml
thezackm: > Multiple definitions could work well if we had a metric attribute that could be used like that `unix:System`.

do we have an example of what this would look like? I can get a PR submitted to the Unix integration to add an `osVersion` field (or something like it); but I'd like to see a working example first so I can make sure I'm following correctly. Thanks!

thezackm:
@IreneMLC 
>Multiple definitions could work well if we had a metric attribute that could be used like that unix:System.

do we have an example of what this would look like? I can get a PR submitted to the Unix integration to add an osVersion field (or something like it); but I'd like to see a working example first so I can make sure I'm following correctly. Thanks!

definitions/infra-unix_host/definition.yml
```diff
@@ -0,0 +1,11 @@
+domain: INFRA
```
srvaroa, line 1: I'm unsure about using INFRA here (specially since there is an INFRA-HOST that arguably has overlap with this one even though we can't reuse that type because it generates custom identifiers). I would default this EXT here (this would mean changing this `domain` attribute, and the name of the folder containing these configs.

thezackm: I don't have a strong opinion either way on this, but I will point out that there's a precedent for using the `INFRA` domain on integrations data:
[Cassandra](https://github.com/newrelic-experimental/entity-synthesis-definitions/tree/main/definitions/infra-cassandranode)
[F5](https://github.com/newrelic-experimental/entity-synthesis-definitions/tree/main/definitions/infra-f5node)
[Kafka](https://github.com/newrelic-experimental/entity-synthesis-definitions/tree/main/definitions/infra-kafkabroker)
[MSSQL](https://github.com/newrelic-experimental/entity-synthesis-definitions/tree/main/definitions/infra-mssqlinstance)
[MySQL](https://github.com/newrelic-experimental/entity-synthesis-definitions/tree/main/definitions/infra-mysqlnode)
[NGINX](https://github.com/newrelic-experimental/entity-synthesis-definitions/tree/main/definitions/infra-nginxserver)
and more...

srvaroa: Yeah, the reason we're a bit unsure about placing this in "INFRA" is because the existing domains have some historical implications tied to internal organization and governance, as well as some technicalities (e.g. many INFRA- entities generally get IDs that are generated internally in NR, and for synthesis we need to ensure that identifiers come directly from telemetry)  Let me get feedback from a couple of internal stakeholders so we can decide for a consistent rule to use.

IreneMLC:
> do we have an example of what this would look like? I can get a PR submitted to the Unix integration to add an osVersion field (or something like it); but I'd like to see a working example first so I can make sure I'm following correctly. Thanks!

What we use as a 'condition' is just another attribute in the telemetry (something like "attributeName":"attributeValue"), but I'm not familiar with the telemetry from that integration so I do not know what the actual value would be, but for example in [this metric](https://chart-embed.service.newrelic.com/herald/40d55138-894d-4fd3-8225-a887939db407?height=1000px&timepicker=true) we could use the value of this attribute to know that it's related to EKS `"k8s.cluster.name": "eks-nsacxp2-b6vv-prgw"`

definitions/infra-unix_host/definition.yml
cianBuckley: Sorry, you're right @IreneMLC This will literally match on anything with a `hostname` attribute - it seems too open. What would be ideal is something like (example): 
```
  conditions:
    - attribute: metricName
      prefix: etcd_
```
Without that, every data point that has `hostname` attribute populated will be organised as a unix host and i don't think we want that

@thezackm Can we see a sample of telemetry that you want to generate entities from?

definitions/infra-unix_host/definition.yml
cianBuckley: For context: There's been an internal decision to match guid formation to the agent reporting it, so we will have 1:1 instrumentation provider to entity type. (Prior we were considering having multiple agents report for a single entity type)
 
For this thread, my interpretation of that is we can push this to match the integration (eg `PROMETHEUS` or `PROM` for char limit) and let `infra` continue being an internal domain.

jfjoly:
@thezackm can we find a more precise condition?
As @cianBuckley [said](https://github.com/newrelic-experimental/entity-synthesis-definitions/pull/14#discussion_r547201200), this will match any telemetry that has `hostname` on it.

definitions/infra-unix_host/definition.yml
srvaroa: i'll close this thread and open a new one just to clear up the PR

definitions/infra-unix_host/definition.yml
srvaroa: Resolving this to clear the PR, I'll make a suggestion for EXT as we're using that pattern for most cases.

srvaroa, line 1:
```suggestion
domain: EXT
```

We'll have to change the directory to reflect this too (I can't do it since the PR is in your fork @thezackm)

definitions/infra-unix_host/definition.yml
srvaroa, line 5:
```suggestion
  identifier: hostname
  conditions:
  - attribute: eventType
     prefix: unixMonitor:
```

cc @jfjoly this is correct right?

srvaroa (commented):
Hi @thezackm I've sent a couple of suggestions that will make this mergeable.  Adding to those, we just need to change the directory to `ext-unix_host` to match the new domain.

Thanks!

definitions/infra-unix_host/definition.yml
```diff
@@ -0,0 +1,14 @@
+domain: EXT
+type: UNIX_HOST
+synthesis:
+  name: hostname
+  identifier: hostname
+  conditions:
+  - attribute: eventType
+     prefix: "unixMonitor:"
```
srvaroa, line 8:
```suggestion
    prefix: "unixMonitor:"
```

srvaroa: This is failing validation, sorry I added an extra char by the looks of it

definitions/ext-unix_host/golden_metrics.yml
```diff
@@ -0,0 +1,26 @@
...
+    facet: hostname
+    eventId: entity.guid
+
+diskUtilizationPercent:
+  title: Disk Utilization Percentage
+  displayAsValue: true
+  query:
+    select: max(percentUsed)
+    from: '`unixMonitor:Disk`'
+    facet: hostname, mountedOn
```
thezackm, line 25: @srvaroa - is multi-facet supported on golden metrics yet?

srvaroa: @naxhh you're more up to date than me on this

srvaroa: @thezackm @naxhh tells me that it's still not supported, but no problem merging this as it is (it will not work straight away, but it's good to give us a first case to verify as soon as we complete that work).

naxhh: @thezackm We support this in the definition but is still not being used in Golden metrics (WIP)

Check: https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/docs/golden_metrics.md#defining-golden-metrics

So when this is implemented this will convert into: `FACET entity.name, hostname, mountedOn`
Since you are using hostname as the `entity.name` you probably want this to look like:

`facet: mountedOn` the facet field is now "extra facets" in the same way we use "where" as extra where aside of `entity.guid IN ....`

hope that makes sense!

thezackm: perfect; I'm good with the changes and ready to go; thanks for all the support!

thezackm: ah, that's right @naxhh ! I forgot about that change since this branch was first made; let me take one last look through the golden and summary to validate things

thezackm: ok; ready to merge

@sschwartzman has mentioned we will want to look at a dashboard later, but that can be added outside of this workflow

naxhh (approved):

PR #20, opened 2020-12-16
Add - as a valid character in type names

IreneMLC:
### Relevant information

Some migrated entity types contain '-' despite it not being on the original spec regex, so I'm adding it now


IreneMLC:
I'm closing this PR for now, there's only one entity that's not following the guideline and I'm tracking down the owner to delete it (maybe we should validate INFRA types though)

PR #21, opened 2020-12-16
Take the composite metrics and golden tags out of the synthesis section

IreneMLC:
### Relevant information

In a previous schema iteration, we moved everything except for the domain and type into the synthesis section, but non-synthesizable entities also have composite metrics and golden tags and they are not really part of the synthesis.


naxhh (approved):

srvaroa (approved):

srvaroa:
This will break some open PRs unfortunately, if we merge it would be nice to push commits to those branches updating them.

IreneMLC:
I've been leaving suggestion edits, should I push the changes directly?

PR #23, opened 2020-12-17
Import golden metrics and tags

naxhh:
### Relevant information

Import golden metrics and golden tags


definitions/infra-apacheserver/definition.yml
```diff
@@ -1,2 +1,6 @@
 domain: INFRA
 type: APACHESERVER
+goldenTags: []
```
IreneMLC, line 3: goldenTags are optional, so if they are empty we could avoid adding the field directly (but I also don't see any harm in this)

IreneMLC (approved):

definitions/infra-apacheserver/definition.yml
naxhh: I did this when they where private definitions to quickly find the ones that don't have any golden tag defined. So devs could scan the project and fix that. We can remove it it's annoying though

IreneMLC: If anyone ever complains it's easy to remove so 👍

cianBuckley (approved):

PR #27, opened 2020-12-21
Add open instrumentation sources

wesrog:
Adds definitions, golden metrics, and summary metrics for OpenTelemetry,
Kamon agent, and micrometer

### Relevant information

First pass at including definitions for open instrumentation sources. Things to note:

1. For OpenTelemetry, since the Metrics spec is still being developed, we are relying on spans for response time, throughput, and error rate. This isn't the case for micrometer and kamon. Once the Metrics spec is GA, we will likely need to change the OTel queries to use Metric instead of Span.

2. I've chosen what I thought is an appropriate domain/type naming structure but I am not attached to it and I'm open to whatever recommendation fits within the suggested entity naming structure.

3. Currently, OI sources are defined as `EXT-SERVICE` with an `instrumentation.provider` tag of `opentelemetry`, `micrometer`, or `kamon-agent`. 

### Checklist

 explanation above.


definitions/oi-kamon-agent/definition.yml
```diff
@@ -0,0 +1,18 @@
+domain: OI
```
wesrog, line 1: I see the validation failed because the domain is too short. Open to recommendations :)

srvaroa: OIT?

srvaroa:
I understand that the end goal here is to be able to define golden/summary metrics targetted to the instrumentation provider (please let me know otherwise).

The approach proposed here has non trivial implications that are worth discussing more deeply.  I'll try to summarize practical implications, and later expand a bit into more theoretical questions regarding the general approach we want to follow in the classification of types.

In practical terms, this PR reclassifies telemetry from one type (EXT-SERVICE) to another (the OI- types). This will mean:

1. Old entities that are no longer reporting telemetry will remain in EXT-SERVICE (until they are TTL'd away)
2. Entities that were reporting telemetry prior to this change and keep doing so afterwards will be duplicated (e.g. there will be an EXT-SERVICE and an OI-MICROMETER entity for the same reporting instance.)  The EXT-SERVICE ones will be TTL'd away after a few days, but both will coexist for a few days.
3. UI will most likely need to be aware of this change because any view that's querying for EXT-SERVICE entities will suddenly stop seeing any (since they are being synthesized as OI-MICROMETER).  In other words, anyone querying the APIs will need to be aware that all three types are "services". 

The latter point is what leads to the a more fundamental discussion, which is about how the way in which we classify entity types affects the ability of our API clients to work with these entities.

The fundamental question is: there are many types of services. Should we have a single supertype "EXT-SERVICE" that captures them all, or specialize entity types for all possible types of SERVICES.

AFAIK main case for having multiple types is that this allows us to specialize things like golden metrics and summary metrics (as is done in this PR.) The problem is how this impacts the ability to work with them through the API. An API client (e.g. the NR UI) will now be unable to request "all services running in my account" (which is an important use case, e.g. "show me all services running in my infrastructure").  The only way to solve it today would be for the client to keep a list saying that OI-KUMON OI-MICROMETER, OI-OPENTELEMETRY and EXT-SERVICE are all services, and maintain this list up to date as new types appear. This is hard to maintain and scale. A new feature that would solve this is having some mechanism to represent supertypes.

The case for a single type is that it solves the problem above by offering explicitly the supertype.  But, today it has the drawback that it's harder to attach golden/summary metrics to the subtypes.  A new feature that would solve this is having the ability to define different golden/summary metrics for a given type based on the `instrumentation.provider` tag.  This requires some development work on our side, but it's doable.
 
My gut feeling is that we should go with the latter as it's more flexible: it allows to have a supertype (SERVICE, as well as the same for others like HOST, CONTAINER, etc.) and allows to specialize the golden / summary metrics with a simple configuration change.  An entity can easily be recategorized by just switching a tag, and this change is lightweight because has no impact on its identity (the type remains the same, as does the GUID).  This flexibility disappears if we specialize more granularly, because it changing the type alters the entity identifier (leading us to problems 1 and 2 described above).  Furthermore, it also requires implementing an external mechanism to represent the supertype (e.g. being able to say "entities of types A, B, C are all of a supertype X). 

@wesrog Do you see a strong case for the granular subtypes other than the ability to define GM/SM's based on the instrumentation provider?

cc. @jasonrclark

cianBuckley (commented):
I'm more or less OK with this.
However, I'd like a more thorough evaluation of the impact of this breaking change before going forth and merging. 
I'd also like to have approval from @jfjoly (or product)

definitions/oi-opentelemetry/definition.yml
```diff
@@ -0,0 +1,20 @@
...
+  name: service.name
+
+  conditions:
+    - attribute: telemetry.sdk.name
+      value: opentelemetry
+
+  tags:
+    - telemetry.sdk.name
+    - telemetry.sdk.language
+    - telemetry.sdk.version
```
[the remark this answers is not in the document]
cianBuckley: example config could be

```
singleValueTags:
   - telemetry.sdk.version
multiValueTags:
   - whateverNeedsMultiValues
```

definitions/oi-kamon-agent/definition.yml
cianBuckley: EXT?

definitions/oi-opentelemetry/definition.yml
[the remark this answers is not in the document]
srvaroa: Yes, that's a feature we'll have to support (both in definition files and in the ingestion pipelines)

jfjoly:
Thanks @wesrog for submitting this! 
Regarding the comments from @srvaroa, I see entity type as a type of objects that not a type of instrumentation. 
Whether you use a New Relic agent, Kamon or Open Telemetry, is it the same object that is monitored? If yes, our goal should be to associate all metrics from all instrumentation with that object. 

Then, for the summary and golden metrics that represents the most important attributes, are the metrics the same but reported by different instrumentation or are each of these instrumentation complimentary? I would expect it to be both. In some cases you will have multiple teams that historically use different instrumentations and in others, one instrumentation goes deeper and get additional metrics in a specific area.  
When the metrics are the same, they will still take different shapes depending on the instrumentation. We may use the `instrumentation.provider` as suggested by @srvaroa  to offer variants according to the different instrumentation sources. 
Note that we probably want `instrumentation.provider` to accept multiple values as @cianBuckley suggested [here](https://github.com/newrelic-experimental/entity-synthesis-definitions/pull/27#discussion_r547311083) if we can have multiple instrumentations for the same object.

wesrog:
> But, today it has the drawback that it's harder to attach golden/summary metrics to the subtypes. A new feature that would solve this is having the ability to define different golden/summary metrics for a given type based on the instrumentation.provider tag. This requires some development work on our side, but it's doable.

@srvaroa Yes, this would solve our issue. We need to be able to define different summary metrics, workload templates, curated UIs, etc based on some defining characteristic, whether that is a domain/type or tag. As long as a customer can experience a full, end-to-end experience with OTel just as if that entity was like a familiar APM entity.

> Do you see other cases for the granular subtypes other than the ability to define GM/SM's based on the instrumentation provider?

No, but @justinfoote or @elucus might?

> I presume we will only ever want the latest version displayed on the entity.

@cianBuckley Yes! Thank you.

> Whether you use a New Relic agent, Kamon or Open Telemetry, is it the same object that is monitored? If yes, our goal should be to associate all metrics from all instrumentation with that object.

@jfjoly Yes, that's a great way of putting it. Agreed.

> Then, for the summary and golden metrics that represents the most important attributes, are the metrics the same but reported by different instrumentation or are each of these instrumentation complimentary? I would expect it to be both.

@jfjoly Metrics might not be the same.

definitions/oi-kamon-agent/definition.yml
```diff
@@ -0,0 +1,18 @@
+domain: EXT
+type: KAMON_AGENT
```
cianBuckley, line 2: Just to start conversation, would this make sense? (mimics our own jvm agent apm|application and browser|application)

```suggestion
domain: KAMON
type: APPLICATION
```

definitions/oi-kamon-agent/definition.yml
```diff
@@ -0,0 +1,18 @@
+domain: EXT
+type: SERVICE
```
wesrog, line 2: @cianBuckley or @srvaroa: For compatibility reasons, I've decided to change these back to just `EXT.SERVICE` instead of `EXT.KAMON`, `EXT.OPENTELEMETRY` et al. Will this cause issues?

wesrog: @srvaroa ping again on this. I am running into validation errors right now, but I wanted to check with you to see if it's ok if we stay with `ext-service` for Open Instrumentation entities in the short-term.

definitions/oi-kamon-agent/definition.yml
```diff
@@ -0,0 +1,18 @@
...
+  conditions:
+    - attribute: instrumentation.provider
+      value: kamon-agent
+
+  tags:
+    - instrumentation.provider
+
+compositeMetrics:
+  goldenMetrics:
+    - ./golden_metrics.yml
```
jonatns, line 16: Looks like the relative path syntax is causing the validation errors
```suggestion
    - golden_metrics.yml
```

definitions/oi-kamon-agent/definition.yml
srvaroa: Hi, sorry I missed this one.  The relative path seems to be the problem yes, I noticed @jonatns noted this below

definitions/oi-kamon-agent/definition.yml
srvaroa: +1

definitions/oi-kamon-agent/definition.yml
```diff
@@ -0,0 +1,18 @@
...
+type: SERVICE
+synthesis:
+  identifier: service.name
+  name: service.name
+
+  conditions:
+    - attribute: instrumentation.provider
+      value: kamon-agent
+
+  tags:
```
srvaroa, line 11: This one doesn't need to be declared explicitly, we always inject it if it's present in the telemetry.

It doesn't hurt to leave it though.

srvaroa:
This one looks fairly close to me.  There is a practical issue though in how these configs are expressed in the definitions. When all rules are meant to match against the same entity type (EXT-SERVICE), the definitions should also go in the definition/ext-service/ path. This is however not yet implemented, so we need to extend the schema to support multiple rules for a single type before merging.

cc @jfjoly we discussed about adding that feature after deciding to go with this solution, but we should decide on priority for implementation so someone can jump on it asap.

srvaroa:
@wesrog these are already supported after changes submitted in a different PR.  Can you confirm that this is the case?

PR #31, opened 2021-01-06
Add rcluster host definition

aschmid13:
### Relevant information

Adding support for rcluster host metrics

### Checklist

 explanation above.


definitions/ext-rclusterhost/definition.yml
```diff
@@ -0,0 +1,24 @@
+domain: EXT
+type: HOST
```
cianBuckley, line 2: This doesn't match the file name like other examples do. Clearly, this is because ext host already exists. As of today, I'm not clear if that is problematic - it depends on two internal decisions. I'll brief them here for transparency

1. we're debating if or not one entity type should have multiple agents (1:1 or N:1 agent:entityType) We should have an answer on this soon (my guess is about a week.. - I'll keep you posted)
2. In the event we permit N:1 agent:entityTypes, we need to decide if this is the format we want to accept (likely it's ok, but need agreement)

definitions/ext-rclusterhost/golden_metrics.yml
```diff
@@ -0,0 +1,15 @@
+rclusterCpuUsage:
+  title: CPU usage (%)
+  query:
+    select: average(rcluster.container.cpu.percentused)*100
+    facet: rfc460Hostname
```
naxhh, line 5: As far as I understand this property is not stored as a tag so it will not exist.

That said since this property is the `synthesis.name` we should be able to facet by `entityName` and have the same result.

We can remove this line since that is the default

aschmid13: Are you referring to the `facet: rfc460Hostname`? Would we need to define it as a tag in definition.yml to use the facet? Since it is the `entityName` it would be a duplicate anyway.

naxhh: Yep sorry I didn't explain myself correctly.

Here you should use `facet: entity.name` since as you said `rfc460Hostname` will be the entity name attribute by default.

And `entity.name` is the default value as you can see in [our docs](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/docs/golden_metrics.md#query-options) so we can remove this line and it will apply correctly.

definitions/ext-rclusterhost/summary_metrics.yml
```diff
@@ -0,0 +1,24 @@
...
+    from: Metric
+    facet: rfc460Hostname
+    eventId: entity.guid
+rclusterStorageUsage:
+  title: Storage usage (%)
+  unit: PERCENTAGE
+  query:
+    select: average(rcluster.container.executor.disk.percentused)*100
+    from: Metric
+    facet: rfc460Hostname
```
naxhh, line 23: As far as i know there's no facet in summary metrics ([docs](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/docs/summary_metrics.md#nrdb-query)). It will use the `eventId` to facet by.

definitions/ext-rclusterhost/golden_metrics.yml
aschmid13: Thanks for the clarification!

definitions/ext-rclusterhost/summary_metrics.yml
aschmid13: updated

definitions/ext-rclusterhost/definition.yml
cianBuckley: So we came to the agreement that N:1 agent:entityTypes is desirable!

As for format, it might make more sense to have all ex-host entity types together. As a suggestion, how about a new agent specific definition file under directory ext-host? eg ext-host/rcluster/{GM/SM/definition}  @srvaroa @IreneMLC opinions?

jfjoly:
This definition is for a specific type of host. I would keep it separated. 
Why not use :
```
domain: EXT
type: RHOST
```

definitions/ext-rclusterhost/definition.yml
srvaroa: Since we want to make this its own type it's fine in a separate directory, but the type in L2 should be `RHOST`.

srvaroa (changes requested):
LGTM, we just need to update the type to `RHOST` in definition.yaml, L2 (I sent a suggested change)

definitions/ext-rclusterhost/definition.yml
srvaroa, line 2:
```suggestion
type: RHOST
```

aschmid13:
The changes make sense to me. I updated with RHOST as the type.

aschmid13 (commented):
Updated to RHOST

srvaroa (approved):
Thanks @aschmid13 !

srvaroa:
@naxhh anything pending from your side?

definitions/ext-rhost/definition.yml
```diff
@@ -0,0 +1,24 @@
+domain: EXT
+type: RHOST
+synthesis:
+  name: rfc460Hostname
+  identifier: rfc460Hostname
```
srvaroa, line 5: One last question: are these values unique per account? what is the max length for them?

aschmid13: It would be a unique IPV4 or FQDN and only report to a single account. The max length for IPV4 is 15 but the FQDN could be longer. I don't think any limits are set aside from the 4096 limit enforced by the Metrics API.

naxhh (approved):
LGTM!

Sorry for my late review

definitions/ext-rhost/definition.yml
srvaroa: It's probably a good idea then to set `encodeIdentifierInGUID: true` in that case (without this there is a chance that the identifier gets trimmed to the first ~50 characters.

aschmid13: Thanks for the callout on that one, I updated the definition to include it.

PR #33, opened 2021-01-11
Validate golden metrics keys contain valid characters

naxhh:
### Relevant information

Make the golden metrics key have a valid format.

The format decided is: a-z A-Z 0-9 _ and from 1 character until 100.
All current definitions are ok with this validator so no further changes are required! \o/

### Checklist

### DIFF
--- a/docs/golden_metrics.md
+++ b/docs/golden_metrics.md
@@ -6,7 +6,8 @@ We allow a maximum of 10 metrics per entityType.
 
 ## Defining golden metrics
 
-Golden metrics are defined in a map where each key is a camel-case name that defines the intention of the metric: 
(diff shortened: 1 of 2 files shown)


srvaroa (approved):

srvaroa:
We probably want to update the documentation though.

naxhh:
I was looking into the docs. Do we want to say that the keys should follow that pattern? I didn't saw any validation specific docs so I wasn't sure if we wanted to add it

IreneMLC:
I think that it can be added to the defining Golden Metrics section https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/docs/golden_metrics.md#defining-golden-metrics we only say that the key needs to be camelcase right now

IreneMLC (approved):

naxhh:
@IreneMLC cool! added docs. let me know if that's good enough.

PR #34, opened 2021-01-12
Support attribute condition

IreneMLC:
### Relevant information

- Support a condition that matches on the presence of a given attribute.
- Mention in the docs that the conditions are not cumulative (only one needs to be fulfilled) and fix the validator to reflect that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Some examples are APPLICATION, HOST or CONTAINER.
   - If you suspect that your identifiers may not fulfil our length requirements, set the optional `encodeIdentifierInGUID` field to true. 
 - The definition needs to provide enough information to differentiate this entity
   from others. It cannot be a subset nor a superset of any existing definition. 
-- If the names of your telemetry attributes are too generic you can define a condition on an attribute that needs to be met for the entity to be synthesized. 
+- If the names of your telemetry attributes are too generic, you can define a condition to match on the value of an attribute. 
   For more information on our supported conditions please refer to the [conditions](#conditions) section. 
 - If you are creating a definition for a `domain` and `type` that already exists we'll need to
  understand your use case, so please provide an explanation in your PR or get in touch with us to discuss it. 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Some examples are APPLICATION, HOST or CONTAINER.
   - If you suspect that your identifiers may not fulfil our length requirements, set the optional `encodeIdentifierInGUID` field to true. 
 - The definition needs to provide enough information to differentiate this entity
  from others. It cannot be a subset nor a superset of any existing definition. If the names of
-  your telemetry attributes are too generic you can define conditions on the value of the field (e.g. `prefix: "eks"`).
+  your telemetry attributes are too generic you can define conditions on the value of the field (e.g. `prefix: "eks"`). The entity will be synthesized if at least one of the specified conditions is met. 
 - If you are creating a definition for a `domain` and `type` that already exists we'll need to
  understand your use case, so please provide an explanation in your PR or get in touch with us to discuss it. 
 - If you are adding composite metrics' files for an entity definition they must be placed inside the same folder and follow the filename format.
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Some examples are APPLICATION, HOST or CONTAINER.
   - 1 to 36 standard ascii characters, excluding control chars (codes: 32-126).
   - If you suspect that your identifiers may not fulfil our length requirements, set the optional `encodeIdentifierInGUID` field to true. 
 - The definition needs to provide enough information to differentiate this entity
- from others. It cannot be a subset nor a superset of any existing definition. 
+  from others. It cannot be a subset nor a superset of any existing definition. 
 - If the names of your telemetry attributes are too generic you can define a condition on an attribute that needs to be met for the entity to be synthesized. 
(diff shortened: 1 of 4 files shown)


docs/example-entity-definition.yml
```diff
@@ -16,13 +16,16 @@ synthesis:
 
   # Conditions that must be met for this entity to be synthesized.
   # Can be added if the attribute used for the identifier is too generic to differentiate this entity from others
+  # These conditions act as an OR clause, the entity will be synthesized if any one of them is fulfilled.
```
srvaroa, line 19: Right now the synthesis library supports only one condition.  Should we update the docs to reflect that? (e.g. the OR won't actually work, you can use only one of those forms).

IreneMLC: My bad, I thought that with the new library we were going to start supporting more than one. I'll update the docs here and add it to the schema.

validator/tools/validate_uniqueness.js
```diff
@@ -8,25 +8,16 @@ function hasDifferentDomainType(entity1, entity2){
     return (entity1.domain !== entity2.domain || entity1.type !== entity2.type)
 }
 
-function hasSameConditions(entity1, entity2){
+function hasConflictingConditions(entity1, entity2){
     let e1conditions = entity1.synthesis.conditions || []
     let e2conditions = entity2.synthesis.conditions || []
 
-    if(isEqual(new Set(e1conditions), new Set(e2conditions))){
+    if(e1conditions.length === 0 || e1conditions.length === 0){
```
naxhh, line 15:
```suggestion
    if(e1conditions.length === 0 || e2conditions.length === 0){
```

naxhh (approved):

validator/tools/validate_uniqueness.js
IreneMLC: thanks!

README.md
```diff
@@ -29,8 +29,9 @@ Some examples are APPLICATION, HOST or CONTAINER.
   - 1 to 36 standard ascii characters, excluding control chars (codes: 32-126).
   - If you suspect that your identifiers may not fulfil our length requirements, set the optional `encodeIdentifierInGUID` field to true. 
 - The definition needs to provide enough information to differentiate this entity
- from others. It cannot be a subset nor a superset of any existing definition. If the names of
-  your telemetry attributes are too generic you can define conditions on the value of the field (e.g. `prefix: "eks"`).
+ from others. It cannot be a subset nor a superset of any existing definition.
```
srvaroa, line 32:
```suggestion
  from others. It cannot be a subset nor a superset of any existing definition. 
```

README.md
```diff
@@ -29,8 +29,9 @@ Some examples are APPLICATION, HOST or CONTAINER.
   - 1 to 36 standard ascii characters, excluding control chars (codes: 32-126).
   - If you suspect that your identifiers may not fulfil our length requirements, set the optional `encodeIdentifierInGUID` field to true. 
 - The definition needs to provide enough information to differentiate this entity
- from others. It cannot be a subset nor a superset of any existing definition. If the names of
-  your telemetry attributes are too generic you can define conditions on the value of the field (e.g. `prefix: "eks"`).
+ from others. It cannot be a subset nor a superset of any existing definition. 
+- If the names of your telemetry attributes are too generic you can define a condition on an attribute that needs to be met for the entity to be synthesized.
```
srvaroa, line 33: Maybe "a condition to match on the value of an attribute" is clearer?

validator/schemas/entity-schema-v1.json
```diff
@@ -93,8 +93,10 @@
         "conditions": {
           "$id": "#/properties/conditions",
           "type": "array",
+          "minItems": 1,
+          "maxItems": 1,
           "title": "Conditions for the entity to be synthesized",
-          "description": "Allows to specify that an attribute's value needs to match a specific value or have a specific prefix.",
+          "description": "Allows to specify that an attribute's value needs to exist, match a specific value or have a specific prefix.",
```
srvaroa, line 99: "that an attribute's value exists, matches a value or a prefix"?

srvaroa, line 99:
```suggestion
          "description": "Allows to specify that an attribute's value exists, matches a prefix or a value.",
```

validator/tools/validate_uniqueness.js
```diff
@@ -35,7 +26,8 @@ const RULES = [
         apply: def => {
             if(def.synthesis != undefined) {
                 const identifier = def.synthesis.identifier
-                if (ENTITY.has(identifier) && hasSameConditions(ENTITY.get(identifier), def) && hasDifferentDomainType(ENTITY.get(identifier), def)) {
+
+                if (ENTITY.has(identifier) && hasConflictingConditions(ENTITY.get(identifier), def) && hasDifferentDomainType(ENTITY.get(identifier), def)) {
```
srvaroa, line 30: The error message needs also updating?

srvaroa (commented):
Minor comments just to make wording more concise, but LGTM

README.md
srvaroa, line 33:
```suggestion
- If the names of your telemetry attributes are too generic, you can define a condition to match on the value of an attribute. 
```

validator/tools/validate_uniqueness.js
IreneMLC: True, thanks!

PR #35, opened 2021-01-12
Remove relative references

IreneMLC:
### Relevant information

- Remove relative references from the definition.yml files
- Add schema validation for the name of the dashboard and composite metric files.

### DIFF
--- a/validator/schemas/entity-schema-v1.json
+++ b/validator/schemas/entity-schema-v1.json
@@ -211,6 +211,10 @@
     "dashboardTemplates": {
       "$id": "#/properties/dashboardTemplates",
       "type": "array",
+      "items": {
+        "type": "string",
+        "pattern": "^([aA-zZ_-])*(dashboard.json)$"
+      },
       "description": "An array dashboard templates",
       "examples": [
         [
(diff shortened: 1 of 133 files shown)


validator/schemas/entity-schema-v1.json
```diff
@@ -211,6 +211,10 @@
     "dashboardTemplates": {
       "$id": "#/properties/dashboardTemplates",
       "type": "array",
+      "items": {
+        "type": "string",
+        "pattern": "^([aA-zZ_-])*(dashboard.json)$"
```
naxhh, line 216:
```suggestion
        "pattern": "^([a-zA-Z_-])*(dashboard.json)$"
```

Is this what you where looking for? 

The current regex will match [, \, ], ^, _, and `

naxhh: Also supper nit but the regex doesn't need the capture groups `^[aA-zZ_-]*dashboard.json$`

naxhh (approved):
LGTM aside of the regex comment.

Nice!

validator/schemas/entity-schema-v1.json
IreneMLC: Thanks! you're right!

IreneMLC:
Already discussed this with @srvaroa offline, so I'll go ahead and merge to unblock ongoing work.

PR #41, opened 2021-01-22
Make conditions mandatory

IreneMLC:
### Relevant information

Having a condition is mandatory for the synthesis. This PR marks it as such on the schema so that it'll be validated, the documentation has also been updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@ so your PR will get merged faster, and you can start enjoying your shiny new ent
 - Each entity definition file must live inside its own folder and both must follow the filename format explained below. 
 - The definition MUST be a valid YAML file.
 - The definition must contain at least the top-level fields `domain` and `type`, along with the fields `name`, `identifier` and `conditions`, located under `synthesis`. 
-We use the `domain`, `type` and `identifier` to assign each entity a Global Unique Identifier (GUID).
+The synthesis section is optional only if you can ensure that the telemetry is being stamped with the entity GUID and tags based on rules defined internally at NewRelic.
+- We use the `domain`, `type` and `identifier` to assign each entity a Global Unique Identifier (GUID).
 - The `domain` must be a value matching `/[A-Z][A-Z0-9_]{2,7}/`. This field is mostly relevant internally for NR. 
 Use EXT by default, although we may advise to use a different value in some cases.                
 - The `type` must be a value matching `/[A-Z][A-Z0-9_]{2,11}/`. This field is meant to identify the type of entity. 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -28,10 +28,9 @@ Some examples are APPLICATION, HOST or CONTAINER.
   - `/[\x20-\x7E]{1,36}/`.
   - 1 to 36 standard ascii characters, excluding control chars (codes: 32-126).
   - If you suspect that your identifiers may not fulfil our length requirements, set the optional `encodeIdentifierInGUID` field to true. 
-- The definition needs to provide enough information to differentiate this entity
-  from others. It cannot be a subset nor a superset of any existing definition. 
-- If the names of your telemetry attributes are too generic, you can define a condition to match on the value of an attribute. 
+- You must also define a condition to match on the value of an attribute. An entity will be synthesized if it matches the provided condition and contains the attribute set for the identifier.
```
srvaroa, line 31: ..and name.

srvaroa (commented):
Minor tweak but LGTM, thanks!

README.md
```diff
@@ -16,7 +16,7 @@ so your PR will get merged faster, and you can start enjoying your shiny new ent
 
 - Each entity definition file must live inside its own folder and both must follow the filename format explained below. 
 - The definition MUST be a valid YAML file.
-- The definition must contain at least the top-level fields `domain` and `type`, along with the fields `name` and `identifier`, located under `synthesis`. 
+- The definition must contain at least the top-level fields `domain` and `type`, along with the fields `name`, `identifier` and `conditions`, located under `synthesis`.
```
naxhh, line 19: Should we rephrase this?

If I'm not wrong is a valid configuration to NOT have synthesis section. right?

For example a domain-type that is not being synthetized using our system but that has golden metrics.

naxhh (approved):

README.md
IreneMLC: It is but I'm not sure if we want to elaborate on that? I'm afraid it might be more confusing for outsider contributors, although maybe we could just say something like `The synthesis section is optional only if you can ensure that the telemetry is being stamped with the entity guid and tags based on rules defined internally at NewRelic` (kinda like [this](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/docs/example-entity-definition.yml#L6)), wdyt?

naxhh: You are right. Whichever you prefer. If we are afraid of confusing contributors we may just leave that detail out. 

I was just thinking that you can open a file now and see that it has no synth section and that may be also confussing.

IreneMLC: You are right, I ended up adding it to the docs, so if anyone's curious at least the information is there.

naxhh (approved):

PR #43, opened 2021-01-25
AWS API Gateway and Workload types

helenapm:
### Relevant information

**definitions/nr1-workload/definition.yml**
Adding some tag keys that appear in workload queries, and might be used by customers to tag their workloads as well. We might refine this list with information provided by the AAI teams.

**definitions/infra-awsapigatewayapi/definition.yml**
**definitions/infra-awsapigatewaystage/definition.yml**
**definitions/infra-awsapigatewayresourcewithmetrics/definition.yml**
I'm adding `aws.region` and `aws.availabilityZone` as golden tags.

**definitions/infra-awsapigatewayapi/golden_metrics.yml**
**definitions/infra-awsapigatewaystage/golden_metrics.yml**
**definitions/infra-awsapigatewayresourcewithmetrics/golden_metrics.yml**
The golden metrics are the same for all entity types, but they have a different `provider`.

I've also added the summary_metrics files that were missing.

### Checklist

* [Yes] I've read the guidelines and understand the acceptance criteria.
* [N/A] The value of the attribute marked as `identifier` will be unique and valid. 
* [N/A] I've confirmed that my entity type wasn't already defined. If it is I'm providing an
 explanation above.


definitions/infra-awsapigatewayresource/definition.yml
```diff
@@ -0,0 +1,3 @@
+domain: INFRA
+type: AWSAPIGATEWAYRESOURCE
+goldenTags: [aws.accountId, aws.region, aws.availabilityZone, aws.Project, aws.Environment, aws.CostCenter]
```
naxhh, line 3: nit: can you provide this in the format:

```
goldenTags:
- tag
- secondTAg
```

Will be easier to avoid conflicts in the future this way

naxhh (commented):
You need to declare:

```
compositeMetrics:
  goldenMetrics:
  - golden_metrics.yml
```

In the definition.yml for the files that have golden metrics. We are looking to remove this requirement moving forward but for now is required

helenapm:
Hi @naxhh , moving the PR to Ready for review at last! Thanks for your help so far!

helenapm:
Sorry for the noise, reviewers, I think I got it right this time (CC @naxhh )

definitions/infra-awsapigatewayresource/definition.yml
```diff
@@ -0,0 +1,11 @@
+domain: INFRA
```
naxhh, line 1: For EP: This one is set to NOT_VISIBLE in our schema. Is one of those from beyond. we are still waiting an answer on what to do with them. should we allow this? Maybe reach to beyond again and take a final stand on this being visible or being deleted?

definitions/infra-awsapigatewayresourcewithmetrics/golden_metrics.yml
```diff
@@ -0,0 +1,24 @@
+requests:
+  title: Requests
+  query:
+    select: sum(provider.count.SampleCount)
+    from: ApiGatewaySample
+    where: provider='ApiGatewayResourceWithMetrics'
+    facet: entityName
+    eventId: entityGuid
+errors4XxAnd5Xx:
```
naxhh, line 9: nit: `4xx` or `4XX` same with `5Xx` this key is used to generate pre computed metrics so we should keep them standard.

naxhh (approved):
Minor details. LGTM overall

Thanks a bunch!

definitions/infra-awsapigatewayresource/definition.yml
srvaroa: +1, this is a public API so I'd avoid adding anything that is still not confirmed

IreneMLC: I would agree on reaching out again, we can point them to this PR to explain that the decision is a blocker?

IreneMLC (approved):
LGTM, although we might want to remove `AWSAPIGATEWAYRESOURCE ` until we know what to do.

definitions/infra-awsapigatewayresource/definition.yml
cianBuckley: easier to be conservative now rather than later.. :) has anyone taken to documenting approaches so we can fully decide what to do here?

cianBuckley (commented):
Apart from visibility discussion, LGTM

PR #44, opened 2021-01-26
Add expiration configuration

IreneMLC:
### Relevant information

Added an optional TTL configuration section, replacing indexStrategy for deletionStrategy, since I think that it's easier to understand. We can also default the deletionFrequency based on the deletionStrategy, so I modified that as well. 

If we do want to keep our original naming convention I can just revert the second commit, so just let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,28 +125,28 @@ dashboardTemplates:
 
 Note that you can define more than one dashboard under the `dashboardTemplates` field. 
 
-#### Entity TTL
+#### Entity expiration 
 
-Entities are deleted by default 8 days after the last time they were indexed or sent out a heartbeat. 
-If this doesn't suit your needs you may also specify your desired behavior under the TTL section of the main `definition.yml` file:  
+By default, entities are automatically deleted if we reach 8 days without receiving any telemetry from them. 
+If this doesn't suit your needs you may also specify your desired `expirationPolicy` and `expirationTime` under the `expiration` section of the main `definition.yml` file:  
 
 ```yaml
-ttl:
-  # The strategy to follow for the entity deletion. Defaults to AUTOMATIC
-  deletionStrategy: AUTOMATIC
-  # The frequency to delete the entity types. Defaults to EIGHT_DAYS
-  deletionFrequency: EIGHT_DAYS
+expiration:
+  # The entity deletion policy. Defaults to AUTOMATIC
+  expirationPolicy: AUTOMATIC
+  # The amount of time without receiving telemetry before an entity is deleted. Defaults to EIGHT_DAYS
+  expirationTime: EIGHT_DAYS
 ```
 
-`deletionStrategy` accepts the values `AUTOMATIC` (default) and `MANUAL`. 
+`expirationPolicy` accepts the values `AUTOMATIC` (default) and `MANUAL`. 
 
-If `AUTOMATIC` is used the entity will be automatically deleted after a certain amount of time, defined by the `deletionFrequency`: 
+If `AUTOMATIC` is used, the entity will be deleted if we don't receive any telemetry for the time defined by `expirationTime`: 
 - `HOURLY`
 - `DAILY`
 - `EIGHT_DAYS` (default)
 - `QUARTERLY`
 
-If `MANUAL` is used, the `deletionFrequency` field must be left empty or set to `MANUAL`. 
+If `MANUAL` is used, the entity will not be deleted based on an `expirationTime` so the field can be left empty. 
 
 ## Testing and validation
 
--- a/README.md
+++ b/README.md
@@ -125,6 +125,29 @@ dashboardTemplates:
 
 Note that you can define more than one dashboard under the `dashboardTemplates` field. 
 
+#### Entity TTL
+
+Entities are deleted by default 8 days after the last time they were indexed or sent out a heartbeat. 
+If this doesn't suit your needs you may also specify your desired behavior under the TTL section of the main `definition.yml` file:  
+
+```yaml
+ttl:
+  # The strategy to follow when indexing the entity type. Defaults to DELETE_BY_QUERY
+  indexStrategy: DELETE_BY_QUERY
+  # The frequency to delete the entity types. Defaults to EIGHT_DAYS
+  deletionFrequency: EIGHT_DAYS
+```
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -125,6 +125,29 @@ dashboardTemplates:
...
+ttl:
+  # The strategy to follow for the entity deletion. Defaults to AUTOMATIC
+  deletionStrategy: AUTOMATIC
+  # The frequency to delete the entity types. Defaults to EIGHT_DAYS
+  deletionFrequency: EIGHT_DAYS
+```
+
+`deletionStrategy` accepts the values `AUTOMATIC` (default) and `MANUAL`. 
+
+If `AUTOMATIC` is used the entity will be automatically deleted after a certain amount of time, defined by the `deletionFrequency`:
```
srvaroa, line 143: Some notes:

* I'd use `expirationPolicy` and `expirationTime`.  Frequency is somewhat confusing for me (I read it as "they get deleted every X days"). Maybe @jfjoly has an opinion here?

* Reword: "will be automatically deleted **if we don't receive any telemetry** for the time defined by `expirationTime`".  Otherwise it reads as if we will delete any entity regardless.

README.md
```diff
@@ -125,6 +125,29 @@ dashboardTemplates:
...
+
+`deletionStrategy` accepts the values `AUTOMATIC` (default) and `MANUAL`. 
+
+If `AUTOMATIC` is used the entity will be automatically deleted after a certain amount of time, defined by the `deletionFrequency`: 
+- `HOURLY`
+- `DAILY`
+- `EIGHT_DAYS` (default)
+- `QUARTERLY`
+
+If `MANUAL` is used, the `deletionFrequency` field must be left empty or set to `MANUAL`.
```
srvaroa, line 149: if the option is not really changing any behaviour it seems simpler to just leave it empty.

docs/example-entity-definition.yml
```diff
@@ -44,3 +44,10 @@ compositeMetrics:
 goldenTags:
   - tagNameA
   - tagNameB
+
+# Defines whether the entities of this type are automatically deleted and with what frequency
+ttl:
```
srvaroa, line 49: Maybe, `expiration` is clearer. (cc @jfjoly thoughts?)

srvaroa:
Thanks! Minor comments!

cianBuckley (commented):
i'd like to specify expiration happens only when there is no telemetry after X period

README.md
```diff
@@ -125,6 +125,29 @@ dashboardTemplates:
...
+  # The strategy to follow for the entity deletion. Defaults to AUTOMATIC
+  deletionStrategy: AUTOMATIC
+  # The frequency to delete the entity types. Defaults to EIGHT_DAYS
+  deletionFrequency: EIGHT_DAYS
+```
+
+`deletionStrategy` accepts the values `AUTOMATIC` (default) and `MANUAL`. 
+
+If `AUTOMATIC` is used the entity will be automatically deleted after a certain amount of time, defined by the `deletionFrequency`: 
+- `HOURLY`
```
gvassallo, line 144: Is this TTL actually used by any entity type? If not, I'd remove it as it would be tricky to support it

IreneMLC: It is used indeed, different entity types have different indexStrategies and deletionFrequencies, the next step after we agree on naming here is to migrate that.

gvassallo: I'm referring to the `HOURLY` one specifically. Sorry for the misunderstanding

IreneMLC: Ah sorry, I misunderstood, I just checked and the only entity using HOURLY is our E2E, so I can remove it from here. Thanks!

gvassallo (approved):

README.md
```diff
@@ -125,6 +125,28 @@ dashboardTemplates:
...
+
+```yaml
+expiration:
+  # The entity deletion policy. Defaults to AUTOMATIC
+  expirationPolicy: AUTOMATIC
+  # The amount of time without receiving telemetry before an entity is deleted. Defaults to EIGHT_DAYS
+  expirationTime: EIGHT_DAYS
+```
+
+`expirationPolicy` accepts the values `AUTOMATIC` (default) and `MANUAL`.
```
naxhh, line 141: Do we even want to expose the manual one? There's no way to manually delete entities from the outside

docs/example-entity-definition.yml
```diff
@@ -44,3 +44,10 @@ compositeMetrics:
 goldenTags:
   - tagNameA
   - tagNameB
+
+# Defines whether the entities of this type are automatically deleted after a certain expirationTime without receiving telemetry
+expiration:
```
naxhh, line 49: thinking out loud. Should this be schema? or configuration? or something vague enough that will allow us to provide more features in the future?

README.md
IreneMLC: There are some entities that use it, like `nr1-workload`,  so we'd have to store that information here. Consulting with Miguel this is something that it seems like do need to support, but it wouldn't make sense for synthesized entities probably, so an option is to leave it in the schema (so that we can add it on the manual ones), and don't document it + add a validation in the emitter that if the synthesis section is not null, the expirationPolicy cant be manual, wdyt?

docs/example-entity-definition.yml
IreneMLC: We could add this under a `configuration` section as you say, or keep adding top level configs that might be unrelated between them (for example, maybe some tag configs in the future?).

README.md
naxhh: Can we simplify this into 

```
expirationTime (EIGHT_DAYS, QUARTERLY, MANUAL)
```

Having one less parameter makes it easy to understand it. And we can map into whatever we need internally.
Sounds it better? or is just worse to do it this way?

docs/example-entity-definition.yml
naxhh: whatever you prefer. I'm trying to think on ways to do as less as breaking changes as possible. But we can live with this I think

IreneMLC: yeah 😅 maybe better safe than sorry, will do

README.md
IreneMLC: That's a good idea, it's easy to understand

naxhh (approved):
great work!

PR #45, opened 2021-01-26
Add synthesis rules for INFRA-CONTAINER

srvaroa:
* Missing aliasing one tag (docker.state should create container.state).
  We have to add this feature to the config files.
* Missing defining a second set of rules for Kubernetes containers,
  which needs adding the multiple-telemetry support feature to these
  files.

This will however allow us to start seeing some synthesized containers.

### Relevant information

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-container/definition.yml
+++ b/definitions/infra-container/definition.yml
@@ -1,6 +1,8 @@
 domain: INFRA
 type: CONTAINER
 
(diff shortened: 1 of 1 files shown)


IreneMLC (approved):
Could we add a comment on the file marking it as incomplete jic?

naxhh (approved):

PR #46, opened 2021-01-26
Migrate vsphere summary metrics

naxhh:
### Relevant information

Move the vsphere summary metrics. These metrics where pointing to wrong entity types in the past and hence why there were not migrated.

### Checklist

 explanation above.


naxhh:
Script was wrong. Opening a new one

PR #47, opened 2021-01-26
Update summary metrics for vSphere, Azure & GCP

naxhh:
### Relevant information

These types where not pointing to the correct entity type. Now they are fixed and hence exported.

We are also updating to new types added today.
From now on all changes for summaries should happen in the public repo

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-azureappserviceplan/definition.yml
+++ b/definitions/infra-azureappserviceplan/definition.yml
@@ -1,2 +1,5 @@
 domain: INFRA
 type: AZUREAPPSERVICEPLAN
+compositeMetrics:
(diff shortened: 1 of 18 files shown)


IreneMLC (approved):

srvaroa (approved):
I took the liberty of updating a typo in the title.

I notice there are more changes than vsphere though, it may make sense to mention them in the title as well so it's kept as easier reference in the log.

naxhh:
@srvaroa yeah. a cheeky PR on the private repo went through when I was doing this :( I'll update it

PR #50, opened 2021-01-27
Provide golden metrics for azure firewall

naxhh:
### Relevant information

For some reason this type was not migrated. So i've manually migrated it.

From now one all GM changes should be done from this repo.

### Checklist

 explanation above.


IreneMLC (approved):
Was it created after the first migration maybe? 🤔

PR #52, opened 2021-01-28
Add alertable configuration

IreneMLC:
### Relevant information

Add the alertable field to the configuration section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,12 +127,15 @@ Note that you can define more than one dashboard under the `dashboardTemplates` 
 
 #### Configurations
 
-In the `configuration` section of the `definition.yml` file you can tweak the entity's behavior. At the moment you can only configure the entity's expiration time. 
+In the `configuration` section of the `definition.yml` file you can tweak the entity's behavior. 
+At the moment we accept two optional configuration parameters `entityExpirationTime` and `alertable`.
 
 ```yaml
 configuration:
   # The amount of time without receiving telemetry before an entity is deleted. Defaults to EIGHT_DAYS
   entityExpirationTime: EIGHT_DAYS
+  # Defines whether the entities of this type should have an alert status associated, defaults to true
+  alertable: true
 ```
 
 By default, entities are automatically deleted if we reach 8 days without receiving any telemetry from them. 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -127,12 +127,15 @@ Note that you can define more than one dashboard under the `dashboardTemplates`
...
 
-In the `configuration` section of the `definition.yml` file you can tweak the entity's behavior. At the moment you can only configure the entity's expiration time. 
+In the `configuration` section of the `definition.yml` file you can tweak the entity's behavior. 
+At the moment we accept two optional configuration parameters `entityExpirationTime` and `alertable`.
 
 ```yaml
 configuration:
   # The amount of time without receiving telemetry before an entity is deleted. Defaults to EIGHT_DAYS
   entityExpirationTime: EIGHT_DAYS
+  # Defines whether the entities of this type should have an alert status associated, defaults to true
```
IreneMLC, line 137: We are actually defaulting to false right now, but almost all entities are `alertable`=true  and for synthesis true probably also makes more sense, wdyt?

docs/example-entity-definition.yml
```diff
@@ -49,3 +49,5 @@ goldenTags:
 configuration:
   # The amount of time without receiving telemetry before an entity is deleted. Defaults to EIGHT_DAYS
   entityExpirationTime: EIGHT_DAYS
+  # Defines whether the entities of this type should have an alert status associated, defaults to true
```
gvassallo, line 52: not sure we want to call this `alert status`, that was the old name for the field. Should be `entity severity` now

IreneMLC: That's true, thanks! the field itself being called `alertable` still make sense right? (cc: @cianBuckley)

gvassallo: `alertable` sounds good to me

README.md
srvaroa: It may be worth adding a brief explanation about how this field will be populated. E.g. "an alertable entity's metadata will include a field `alertSeverity` that is updated when the telemetry associated to this entity breaks an alerting condition."

naxhh (approved):

README.md
srvaroa: ping! (I can help out if you're busy!)

IreneMLC:
I'll go ahead and merge this but I'll do another PR updating the docs so I can add more information on the `alertable` if we want @srvaroa

PR #59, opened 2021-02-04
Update golden metrics units for APM

jfjoly:
- Multiplied latency by 1000 to show in milliseconds.
- Added the unit to throughput: calls per minutes.

### Relevant information

### Checklist

 explanation above.

### DIFF
--- a/definitions/apm-application/golden_metrics.yml
+++ b/definitions/apm-application/golden_metrics.yml
@@ -1,11 +1,11 @@
 responseTimeMs:
   title: Response time (ms)
   query:
-    select: average(newrelic.timeslice.value)
+    select: average(newrelic.timeslice.value) * 1000
     where: metricTimesliceName in ('HttpDispatcher', 'OtherTransaction/all')
     facet: appName
 throughput:
-  title: Throughput
+  title: Throughput (calls per minutes)
   query:
     select: count(newrelic.timeslice.value)
     where: metricTimesliceName in ('HttpDispatcher', 'OtherTransaction/all')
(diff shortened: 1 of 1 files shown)


srvaroa:
I think a better approach for this would be that the backend takes the `title` field of every GM, and adds that automatically in an `as` for every query.   Benefits:

- Avoid repeating the same text over and over again in every single file.
- Avoid inconsistencies (e.g. `title` being "Throughput (req/s)" and the `as` being "Requests per second").

When we start adding multiple telemetry sources for the same entity type, both points above are exacerbated. E.g. we may have N GM definitions inside each type. So it's likely that for the same metric (e.g. `Throughput (req/s)`) each telemetry source writes their own label.  This will be bad in the UI as the goal there is to render 1 entity type consistently, but now different entities of that type will label the same metric differently.

The change in the backend for this should be relatively easy IMO (as noted, it's just adding the $title to the query).

gvassallo:
@srvaroa I'm not sure using the title is the best approach though. The title and the name of the metric are 2 different things in the graph. Using the same for both would result in duplicated information. I'd rather use the "humanized" `name` - e.g. `requestPerSecond` becomes "Request Per Second" - instead of the title.
And/Or we could add another property to the golden metric definition to allow overriding it.

cianBuckley (approved):

cianBuckley:
LGTM

@srvaroa @gvassallo short of giving units to nrdb, perhaps an optional enum to opt into a unit type could assist.. but any and 
all solutions here is likely going to be a can of worms.. we can discuss further though.

srvaroa:
Solved in #76

naxhh:
@srvaroa we still need the `* 1000` in this pr

PR #60, opened 2021-02-04
Update summary metrics schema

IreneMLC:
### Relevant information

Add the eventObjectId enum so that it gets validated if used, also fix an error with the key metrics.

### DIFF
--- a/validator/schemas/summary-metrics-schema-v1.json
+++ b/validator/schemas/summary-metrics-schema-v1.json
@@ -110,9 +110,14 @@
             "examples": [
               "entity.guid"
             ]
+          },
+          "eventObjectId": {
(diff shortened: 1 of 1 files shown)


naxhh (approved):
<3

validator/schemas/summary-metrics-schema-v1.json
```diff
@@ -128,7 +133,7 @@
           "key"
         ],
         "properties": {
-          "select": {
+          "key": {
```
srvaroa, line 136: Does this require modifying existing files?

IreneMLC: No, luckily all existing files are valid, so this change has no effect, but before we were basically not validating tags metrics.

PR #74, opened 2021-02-09
add support for multivalue/singlevalue tags

cianBuckley:
### Relevant information

Adds support for single value tags

### Checklist

 explanation above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,13 @@ During synthesis, the tags will be created using the attribute name as key and i
 
 If some of the tags' attributes are not present on the telemetry message received, the entity will still be synthesized with the available tags (if any).
 
+```yaml
+  tags:
+    - attribute: attributeNameB
+      multiValue: false # declare if the value for this tag should be a list of string values (accumulated) or a single value (replaced). defaults to true
+    - attribute: attributeNameC
+```
+
 #### Golden tags
 
 The `goldenTags` field accepts an array of tag-keys which are considered the most important for the entity's DOMAIN and TYPE, 
--- a/definitions/infra-container/definition.yml
+++ b/definitions/infra-container/definition.yml
@@ -12,12 +12,14 @@ synthesis:
     - attribute: docker.containerId
 
   tags:
-    - container.state
-    - docker.state
-    - docker.containerId
-    - newrelic.integrationName
-    - newrelic.integrationVersion
-    - newrelic.agentVersion
+    - attribute: container.state
+      mutliValue: false
+    - attribute: docker.state
+      mutliValue: false
+    - attribute: docker.containerId
+    - attribute: newrelic.integrationName
+    - attribute: newrelic.integrationVersion
+    - attribute: newrelic.agentVersion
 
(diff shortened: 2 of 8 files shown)


definitions/infra-container/definition.yml
```diff
@@ -12,12 +12,14 @@ synthesis:
...
-    - container.state
-    - docker.state
-    - docker.containerId
-    - newrelic.integrationName
-    - newrelic.integrationVersion
-    - newrelic.agentVersion
+    - attribute: container.state
+      mutliValue: false
+    - attribute: docker.state
+      mutliValue: false
```
cianBuckley, line 18: I think these are solid use cases for it, but just want to highlight these lines

README.md
```diff
@@ -95,6 +95,13 @@ During synthesis, the tags will be created using the attribute name as key and i
 
 If some of the tags' attributes are not present on the telemetry message received, the entity will still be synthesized with the available tags (if any).
 
+```yaml
+  tags:
+    - attribute: attributeNameB
+      multiValue: false # declare if the value for this tag should be a list of string values (accumulated) or a single value (replaced). defaults to true
```
srvaroa, line 101: @jfjoly from your perspective does this sound meaningful, or would append / replace work better?

definitions/infra-container/definition.yml
srvaroa: s/mutliValue/multiValue

README.md
```diff
@@ -95,6 +95,13 @@ During synthesis, the tags will be created using the attribute name as key and i
 
 If some of the tags' attributes are not present on the telemetry message received, the entity will still be synthesized with the available tags (if any).
 
+```yaml
+  tags:
+    - attribute: attributeNameB
```
srvaroa, line 100: Is is worth adding the attribute keyword? if we just leave `attributeNameB` it would also work, but being more concise?

cianBuckley: its an object so i think it needs it

srvaroa: This is how we do it in golden metrics:

https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/definitions/infra-awsapigatewayapi/golden_metrics.yml#L9

Each of these are objects, but the name of the metric functions as identifier:

image

cianBuckley: Personally, I prefer it the way it is. From a clients perspective I don't think either way has much impact.
from an internal perspective, IMO it it is easier to operate on in code like such:
```
class Tag(attribute, multiValue = true)
List<Tag>
```
as apposed to 
```
class TagProperties(multiValue=true)
List<Map<String, TagProperties>
```

IreneMLC (approved):
LGTM

README.md
srvaroa: @jfjoly any preference?

naxhh:
Im closing this one and I'll open a new PR. We are going with a different format and there are other PR's that will force us to resolve conflicts so it's easier for me to open a new one :D

PR #76, opened 2021-02-10
Add titles to golden metrics

cirne:
Use `AS ...` to specify a more human readable title for golden metrics (currently showing up as "Binop" and "newrelic.timeslice.value" in chart legends)

### Relevant information

### Checklist

 explanation above.

### DIFF
--- a/definitions/apm-application/golden_metrics.yml
+++ b/definitions/apm-application/golden_metrics.yml
@@ -1,13 +1,13 @@
 responseTimeMs:
   title: Response time (ms)
   query:
-    select: average(newrelic.timeslice.value)
+    select: average(newrelic.timeslice.value) AS 'Response time (ms)'
     where: metricTimesliceName in ('HttpDispatcher', 'OtherTransaction/all')
     facet: appName
 throughput:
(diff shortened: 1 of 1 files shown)


srvaroa (approved):
Thanks @cirne!

FYI: we were just discussing about this because it feels like forcing the `as` clause in all the golden metrics has two drawbacks:

* Potential inconsistencies: as we'll be supporting soon multiple sources of telemetry for the same entity type, this will require declaring multiple queries for each Golden Metric.  With the `as` approach, there is a potential for making the titles inconsistent in different queries for the same metric (e.g. "throughput" vs. "requests / s") making the UI inconsistent.
* Poor UX (users have to write `as` clauses all over the place.

We actually have this info already in the config files. E.g. In L1, L7 and L13 we already know what each of these metrics are. So we are discussing with @jfjoly whether it'd be better to automatically inject the metric name as an `as` clause at query time.  We'll run the text through a humanizer (e.r. errorRate will become "Error rate").

It'd be interesting to get your thoughts on that approach.  I'll merge this now so it stats reflecting in the UI.

PR #79, opened 2021-02-11
Hotfix remove hardcoded summaries

naxhh:
### Relevant information

There are a few types that have special requirements in the summary metrics.
Currently we can't define those requirements with the rules we have in the public repo.
This PR is to ensure we don't allow to define summary metrics for those types.

Looking into the future we should start moving them into this repo. But for now we should ensure there are no definitions of them

### Checklist

 explanation above.


validator/tools/validate_uniqueness.js
```diff
@@ -34,6 +34,46 @@ const RULES = [
                 ENTITY.set(identifier, def)
             }
         }
+    },
+    {
+        name: 'Summary metrics for this type is not allowed',
```
IreneMLC, line 39:
```suggestion
        name: 'Summary metrics for this type are not allowed',
```

IreneMLC (approved):
Nice! 👏

definitions/infra-container/definition.yml
```diff
@@ -26,5 +26,3 @@ goldenTags:
 compositeMetrics:
   goldenMetrics:
   - golden_metrics.yml
-  summaryMetrics:
```
cianBuckley, line 29: Might be worth adding a comment here explaining why we're preventing them from being (re-)added until we fix

PR #81, opened 2021-02-11
Update golden metrics for INFRA-CONTAINER

areina:
### Relevant information

- Update queries to use sample events instead of the new Metric event
type. There are still some services consuming these queries that can't
handle the new format.
- Remove the `as` clause in the select fields.
- Fix the queries to use the same kind of metrics between different data
sources.

@naxhh @IreneMLC do we support the double event types syntax, ex: `from: ContainerSample, K8sContainerSample` ?

That's supported by nrql and nrdb, I hope we also support it here.

Thank you

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-container/golden_metrics.yml
+++ b/definitions/infra-container/golden_metrics.yml
@@ -1,20 +1,21 @@
 cpuUsage:
   title: CPU usage (cores)
   query:
-    select: max(docker.container.cpuUsedCores) or max(k8s.container.cpuUsedCores)
-      as 'CPU used cores'
-cpuUtilization:
-  title: CPU Utilization (%)
-  query:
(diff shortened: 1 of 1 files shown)


cianBuckley (approved):

cianBuckley:
Unfortunately, there are performance/scaling concerns related to the double event table scoping in these queries. We need to make some internal changes before being able to merge this through. We will keep you posted!

naxhh:
Closing this one since we will not merge it as-is.

I would suggest using the [roll-up](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/docs/golden_metrics.md#roll-up-entities) functionality if we still want to provide this feature.

PR #83, opened 2021-02-13
Create entity for a Fastly Datacenter (aka POP)

nr-jbiggley:
### Relevant information

Parsing Logs-to-Metrics and then synthesizing the metrics to create an entity for a Fastly datacenter (POP)

### Checklist

 explanation above.


nr-jbiggley (commented):
Needed to change the L2M rule name

jfjoly:
@nr-jbiggley the `-` is not accepted in the name of the type. Also the length for the name of the type is limited to 11.
Here is the regexp that it needs to match `/[A-Z][A-Z0-9_]{2,11}/` (more details in the README of this repo). You can use https://regexr.com/ to verify that it's matching the regexp.

definitions/ext-fastly-datacenter/definition.yml
```diff
@@ -0,0 +1,33 @@
+domain: EXT
+# Derived from the Fastly log ingestion, a datacenter is a Point of Presence
+# https://docs.fastly.com/en/guides/fastly-pop-locations
+type: FASTLY-DATACENTER
```
srvaroa, line 4:
```suggestion
type: FASTLY_DATACENTER
```

`-` is a reserved character

definitions/ext-fastly-datacenter/definition.yml
```diff
@@ -0,0 +1,33 @@
...
+
+synthesis:
+  name: fastly-datacenter
```
srvaroa, line 7: Some questions: 

- Are `name` and `identifier` meant to be the same (they use `-` and `_`)
- I assume the `fastly-datacenter` property in this telemetry will be considered unique for a given account, is that right?
- Do you know how long are `fastly-datacenter` values? (this is relevant in order to know if we should encode the identifiers, they would remain untouched in the name.

srvaroa (changes requested):
Minor changes and questions, but LGTM.

Thanks @nr-jbiggley !

definitions/ext-fastly-datacenter/definition.yml
nr-jbiggley: @srvaroa The fastly-datacenter property is unique for a given account. Each Datacenter has a 3 character code that is expressed by Fastly.

nr-jbiggley (commented):
Looks solid to me. Thanks for the help @jfjoly & @srvaroa. The first one is always the hardest :)

definitions/ext-fastly-datacenter/definition.yml
srvaroa: Fantastic, thanks!

srvaroa:
Sorry @nr-jbiggley I missed this in my previous review: the directory should match exactly this format "$domain-$type" in lowercase.  Can you rename it to `ext-fastly_pop`? I'll merge right after that's done.

srvaroa (approved):

srvaroa:
thanks @nr-jbiggley !

PR #85, opened 2021-02-16
Add support for hidden summary metrics

IreneMLC:
### Relevant information

Add support for hidden summary metrics. This allows us to move here some missing SMs and removed them from the list of entities with private definitions.

- INFRA-KUBERNETESCLUSTER was not migrated since it contains the deprecated ENTITY_NAME


definitions/infra-container/summary_metrics.yml
```diff
@@ -0,0 +1,103 @@
...
+cpuUsedCoresK8s:
+  query:
+    eventId: entityGuid
+    from: K8sContainerSample
+    select: average(cpuUsedCores)
+  title: CPU usage (cores)
+  unit: COUNT
+  hidden: true
+cpuUsedCores:
+  derive: "@cpuUsedCoresContainer || @cpuUsedCoresK8s \n"
```
naxhh, line 22: \n ?

Something weird happened in this file. Check that there are several quotes with line breaks.

definitions/infra-host/summary_metrics.yml
```diff
@@ -0,0 +1,135 @@
...
+    from: GcpVirtualMachineSample
+    select: average(instance.cpu.Utilization)
+    eventId: entityGuid
+  hidden: true
+cpuUtilizationPercent:
+  unit: PERCENTAGE
+  derive: '@cpuUtilizationPercentEc2 || @cpuUtilizationPercentAzure || @cpuUtilizationPercentGcp
+    || @cpuUtilizationPercentHost
+
+    '
```
naxhh, line 38: same in this file

definitions/mobile-application/summary_metrics.yml
```diff
@@ -0,0 +1,91 @@
...
+    select: count(*)
+    where: errorType = 'NetworkFailure'
+    eventId: appId
+    eventObjectId: DOMAIN_ID
+  hidden: true
+totalNetworkCount:
+  unit: COUNT
+  derive: '@networkFailureCount + @httpErrorCount + @httpRequestCount
+
+    '
```
naxhh, line 66: this one also :(

validator/schemas/summary-metrics-schema-v1.json
```diff
@@ -114,7 +135,7 @@
           "eventObjectId": {
             "$id": "#/properties/query/properties/eventObjectId",
             "type": "string",
-            "enum": ["DOMAIN_IDS","ENTITY_GUIDS"]
+            "enum": ["DOMAIN_ID","ENTITY_GUIDS"]
```
naxhh, line 138: This is `DOMAIN_IDS` in the lib.

Checking around golden metrics also uses DOMAIN_IDS. And summary metrics uses DOMAIN_ID. This is probably my fault that I pluralized it in GM...

I would say to keep the plural and do the conversion in summary metrics (should already be that way)
But check the imported definitions to ensure they bring DOMAIN_IDS not DOMAIN_ID

validator/tools/validate_rules.js
```diff
@@ -40,18 +40,7 @@ const RULES = [
...
-                "INFRA-CONTAINER",
-                "INFRA-COUCHBASENODE",
-                "INFRA-F5NODE",
-                "INFRA-F5POOL",
-                "INFRA-F5POOLMEMEBER",
-                "INFRA-F5VIRTUALSERVER",
-                "INFRA-F5VIRTUALSERVER",
-                "INFRA-HOST",
                 "INFRA-KUBERNETESCLUSTER",
-                "INFRA-RABBITMQNODE",
```
naxhh, line 53: yay! why we can't expose K8s?

naxhh (changes requested):
The DOMAIN_IDS change could break things.

Aside of that minor notes
nice work!

definitions/infra-container/summary_metrics.yml
IreneMLC: ah, the migration script doing weird things, sorry!

validator/schemas/summary-metrics-schema-v1.json
IreneMLC: Right now SM will fail if we try to pass the DOMAIN_IDS plural I think, I can change that before merging this then. There are no SM in this repo that use the property btw, this is the first one.

validator/tools/validate_rules.js
IreneMLC: Because it has the unsupported `ENTITY_NAME` 😞

naxhh: Oh true... We can keep that for a follow up I think. Because we will need the same in GM

definitions/infra-host/summary_metrics.yml
```diff
@@ -0,0 +1,124 @@
...
+  query:
+    from: ComputeSample
+    select: average(provider.networkOutBytes.Average)
+    eventId: entityGuid
+  hidden: true
+networkTrafficEc2:
+  unit: BYTES_PER_SECOND
+  derive: '(@networkInEc2 + @networkOutEc2) / 60
+
+    '
```
naxhh, line 84: found another one :D

IreneMLC: 🤦‍♀️ I'm blind

naxhh (approved):
great job!

PR #87, opened 2021-02-18
Add OpenTelemetry Kubernetes tags.

fryckbos:
### Relevant information

Added K8s tags for OpenTelemetry services as defined here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/k8s.md

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-service/definition.yml
+++ b/definitions/ext-service/definition.yml
@@ -8,3 +8,9 @@ synthesis:
 
   conditions:
     - attribute: service.name
+
+  tags:
+    - k8s.cluster.name
+    - k8s.namespace.name
+    - k8s.pod.name
+    - k8s.node.name


definitions/ext-service/definition.yml
```diff
@@ -8,3 +8,9 @@ synthesis:
 
   conditions:
     - attribute: service.name
+
+  tags:
```
srvaroa, line 12: @fryckbos you're expecting these to be aliased to k8s.clusterName etc. right?

We don't have this supported in the definition files yet (it is in the backend) but I will add a task to add it if that's the case.

IreneMLC: We've added support for this, you can add it like [this](https://github.com/newrelic-experimental/entity-synthesis-definitions#tags):

```
tags:
    k8s.cluster.name:
      entityTagName: k8s.clusterName # preferredTagName
```

naxhh:
@fryckbos is this one still relevant?

naxhh:
Closing. Feel free to reopen if needed

PR #88, opened 2021-02-18
Tags definitions are now a map instead of a list

naxhh:
### Relevant information

In future we will need to add behavior to the tags. For example single value tags as this PR introduces. Aliases for tags, etc..

In order to do that we need to allow a broad configuration to the tags instead of just a simple string.

### Checklist

 explanation above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,13 @@ During synthesis, the tags will be created using the attribute name as key and i
 
 If some of the tags' attributes are not present on the telemetry message received, the entity will still be synthesized with the available tags (if any).
 
+```yaml
+  tags:
+    attributeNameB:
+      multiValue: false # declare if the value for this tag should be a list of string values (accumulated) or a single value (replaced). defaults to true
+    attributeNameC:
+```
+
 #### Golden tags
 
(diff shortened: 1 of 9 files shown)


srvaroa (approved):

README.md
```diff
@@ -95,6 +95,13 @@ During synthesis, the tags will be created using the attribute name as key and i
 
 If some of the tags' attributes are not present on the telemetry message received, the entity will still be synthesized with the available tags (if any).
 
+```yaml
+  tags:
+    attributeNameB:
+      multiValue: false # declare if the value for this tag should be a list of string values (accumulated) or a single value (replaced). defaults to true
```
srvaroa, line 101: I would explain this more in detail as it can be quite confusing, and add an example for when you want to set this to false. E.g. the case for containers is good: "By default, an entity tag will contain all the values seen for this attribute in the telemetry (`multivalue: true`).  When `multiValue` is `false`, then the entity tag value will be overriden with the latest value of the corresponding attribute seen in the telemetry.  An example can be.. )

srvaroa:
LGTM although I'd splash on the documentation a bit more.

README.md
naxhh: Agreed. I'll do another PR improving docs.

PR #89, opened 2021-02-18
Update docs to reflect guid length changes

naxhh:
### Relevant information

As part of synthesis work we increased the guid limits. Reflect that into the docs

### Checklist

 explanation above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,14 +19,14 @@ so your PR will get merged faster, and you can start enjoying your shiny new ent
 - The definition must contain at least the top-level fields `domain` and `type`, along with the fields `name`, `identifier` and `conditions`, located under `synthesis`. 
 The synthesis section is optional only if you can ensure that the telemetry is being stamped with the entity GUID and tags based on rules defined internally at NewRelic.
 - We use the `domain`, `type` and `identifier` to assign each entity a Global Unique Identifier (GUID).
-- The `domain` must be a value matching `/[A-Z][A-Z0-9_]{2,7}/`. This field is mostly relevant internally for NR. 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -19,14 +19,14 @@ so your PR will get merged faster, and you can start enjoying your shiny new ent
 - The definition must contain at least the top-level fields `domain` and `type`, along with the fields `name`, `identifier` and `conditions`, located under `synthesis`. 
 The synthesis section is optional only if you can ensure that the telemetry is being stamped with the entity GUID and tags based on rules defined internally at NewRelic.
 - We use the `domain`, `type` and `identifier` to assign each entity a Global Unique Identifier (GUID).
-- The `domain` must be a value matching `/[A-Z][A-Z0-9_]{2,7}/`. This field is mostly relevant internally for NR. 
+- The `domain` must be a value matching `/[A-Z][A-Z0-9_]{2,15}/`. This field is mostly relevant internally for NR.
```
IreneMLC, line 22: The docs don't have the same limits as the rules (14, 49) which are correct, also, we might want to mention the max guid length here?

naxhh: Yep this ones are wrong! I did the same mistake twice :D 

About the max length. How important is that here? no one is directly creating the guids so that constraint doesn't apply to them, right?

IreneMLC: 😄 the magic of repeated info. Yeah, you are right, I was thinking of it in terms of the conversation we had on the specs repo, but it makes more sense there, this looks nice

IreneMLC (approved):

PR #90, opened 2021-02-19
Improve docs on multiValue use case

naxhh:
### Relevant information

Update docs as discussed in previous PR to make clearer the intent of multiValue config property.

### Checklist

 explanation above.


README.md
```diff
@@ -95,13 +95,19 @@ During synthesis, the tags will be created using the attribute name as key and i
 
 If some of the tags' attributes are not present on the telemetry message received, the entity will still be synthesized with the available tags (if any).
 
+By default an entity tag will contain all the values seen for this attribute in the telemetry.
+If you want to override all the values with the new one you can configure `multiValue: false` for that specific tag
```
IreneMLC, line 99:
```suggestion
If you want to override all the values with the new one you can configure `multiValue: false` for that specific tag.
```

README.md
```diff
@@ -95,13 +95,19 @@ During synthesis, the tags will be created using the attribute name as key and i
...
+
 ```yaml
   tags:
     attributeNameB:
-      multiValue: false # declare if the value for this tag should be a list of string values (accumulated) or a single value (replaced). defaults to true
+      multiValue: false
     attributeNameC:
 ```
 
+An example for this is `INFRA-CONTAINER` where the tag `container.state` will always display the last state (`RUNNING`, `STOPPED`, etc..) instead of a list of all the states the entity has gone trough
```
IreneMLC, line 108:
```suggestion
An example of this is `INFRA-CONTAINER` where the tag `container.state` will always display the last state (`RUNNING`, `STOPPED`, etc..) instead of a list of all the states the entity has gone through.
```

IreneMLC (approved):
Just a few nitpicks.

srvaroa (approved):

README.md
```diff
@@ -95,13 +95,19 @@ During synthesis, the tags will be created using the attribute name as key and i
...
+
 ```yaml
   tags:
     attributeNameB:
-      multiValue: false # declare if the value for this tag should be a list of string values (accumulated) or a single value (replaced). defaults to true
+      multiValue: false
     attributeNameC:
 ```
 
+An example of this is `INFRA-CONTAINER` where the tag `container.state` will always display the last state (`RUNNING`, `STOPPED`, etc..) instead of a list of all the states the entity has gone through.
```
srvaroa, line 108: nit: the values are actually lowercase!

PR #91, opened 2021-02-19
Update SM&GM for INFRA-AZUREAPIMANAGEMENTSERVICE

areina:
### Relevant information

Fix the title of a metric which value is in milliseconds.

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-azureapimanagementservice/summary_metrics.yml
+++ b/definitions/infra-azureapimanagementservice/summary_metrics.yml
@@ -10,8 +10,8 @@ requests:
     eventId: entityGuid
     select: sum(totalRequests.Total)
     from: AzureApiManagementServiceSample
-averageResponseTimeS:
-  title: Average response time (s)
+averageResponseTime:
+  title: Average response time (ms)
   unit: SECONDS
   query:
     select: average(durationMilliseconds.Average)
--- a/definitions/infra-azureapimanagementservice/summary_metrics.yml
+++ b/definitions/infra-azureapimanagementservice/summary_metrics.yml
@@ -11,10 +11,10 @@ requests:
     select: sum(totalRequests.Total)
     from: AzureApiManagementServiceSample
 averageResponseTime:
-  title: Average response time (ms)
+  title: Average response time
   unit: SECONDS
   query:
-    select: average(durationMilliseconds.Average)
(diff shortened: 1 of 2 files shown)


definitions/infra-azureapimanagementservice/summary_metrics.yml
```diff
@@ -11,7 +11,7 @@ requests:
     select: sum(totalRequests.Total)
     from: AzureApiManagementServiceSample
 averageResponseTimeS:
-  title: Average response time (s)
+  title: Average response time (ms)
   unit: SECONDS
```
areina, line 15: It isn't clear to me what we should do with the unit when the metric is in milliseconds. Should we divide the value by 1000 in the query and leave the metric in seconds?

naxhh: Yes we should. In fact the UI adapts to whatever it seems fit.

So if the value is 0.05s it will probably show 5ms which makes me thing that the unit in the title is probably not a good thing.. but that's probably for another PR/discussion

srvaroa: My knee-jerk reaction is that it'd be nice for all entity types to report times in the same unit, but at the same time I can see this being impractical (e.g. one entity type having response times where the reasonable unit can be ms, another entity type having a different time metric where ms is too small).  For now it seems ok that we optimize locally to each type. WDYT?

srvaroa (approved):

definitions/infra-azureapimanagementservice/summary_metrics.yml
areina: What should we put as unit in that case? I don't think we expose `MILLISECONDS` as a unit and having to specify in the title seems pretty hacky.

gvassallo: it won't always be displayed in ms in the UI as it changes based on the value. e.g. 1000 ms might become 1s. I would not specify the unit in the title

areina:
Hey folks, I've updated the PR addressing your feedback and with the last version of the main branch. Would you mind taking a look again, please? Thank you!

naxhh (approved):

PR #93, opened 2021-02-22
Dont allow GM & Golden Tags for INFRA-K8SCluster

naxhh:
### Relevant information

INFRA-K8SContainers is using a feature that we want to deprecate.
Because of that we will keep the definition as private. All changes to it should go via tickets

### Checklist

 explanation above.


naxhh:
I'll merge this one when internal changes are ready

IreneMLC (approved):
Thanks!

PR #100, opened 2021-02-24
Add rabbitmqnode SM and remove from disallow-list

IreneMLC:
### Relevant information

Confirmed with the owner that they are OK with using these metrics instead of the ones we had defined internally.


naxhh (approved):
🎉

PR #109, opened 2021-03-05
New definitions format

naxhh:
### Relevant information

This Pr brings the new definitions format allowed with the new lib.

The major change is allowing `queries` in GM & SM. But since we still allow `query` nothing needs to be updated.

A breaking change is that now dashboards are a list of objects, so the script updates that.
Finally `facet` in a GM query is now optional. And we move the entityName field to `eventName`

Schemas updated acordingly

NOTE: we are missing a new change where synthesis allow a lost of rules so this is ready to review but not merge.
NOTE: This should be merged after we update some internal services.

### Checklist

 explanation above.


README.md
```diff
@@ -132,12 +132,23 @@ You can create a dashboard with the NewRelic interface and export it to JSON for
...
 Note that you can define more than one dashboard under the `dashboardTemplates` field. 
 
+If your entity has different sources you can also provide a rule to show the dashboard only if the condition matches.
+The condition is that the entity has the tag `instrumentation.provider` with the value specified in the `tag` field when tag is used we also require a `title`:
+
+```yaml
+dashboardTemplates:
+ - title: Prometheus only dashboard
+   template: ./prometheus_dashboard.json
+   tag: prometheus
```
srvaroa, line 148: I'd find this easier to parse like this:

```
dashboardTemplates:
  prometheus:
    title:
    template: 

As it's clearer that "prometheus" functions as the identifier, and we don't need to leak the detail that this is a tag in the configuration (if we stay with `tag: ..` I'd use a different name, e.g. `instrumentation_provider: ` or `telemetry_source`)

File: README.md [hunk 2]
```diff
@@ -132,12 +132,23 @@ You can create a dashboard with the NewRelic interface and export it to JSON for
...
 ```yaml
-dashboardTemplates: 
- - ./<dashboardName>.json
+dashboardTemplates:
+ - template: ./<dashboardName>.json
 ```
 
 Note that you can define more than one dashboard under the `dashboardTemplates` field. 
 
+If your entity has different sources you can also provide a rule to show the dashboard only if the condition matches.
```
srvaroa, line 141: I'd reword to something like this:

```
When an entity type has several telemetry sources, you can provide a different dashboard for each one.  To add a dashboard specific for a telemetry source, add a block in the `dashboardTemplates` section like this:
```

Not completed as it depends on how the yaml looks like!

definitions/apm-application/golden_metrics.yml
```diff
@@ -3,16 +3,16 @@ responseTimeMs:
   query:
     select: average(newrelic.timeslice.value) * 1000 AS 'Response time (ms)'
     where: metricTimesliceName in ('HttpDispatcher', 'OtherTransaction/all')
-    facet: appName
+    eventName: appName
```
srvaroa, line 6: Is this really an event name? (e.g. do events have a name?)

naxhh: this is an eventName as envetId is an id of the event :)

don't love the names of those but we keep consistency since they are exposed in some apis

docs/golden_metrics.md
```diff
@@ -47,7 +48,29 @@ memoryUsage:
...
+      select: average(nrField)
+      from: NewRelicSample
+```
+
+In this example `prometheus` and `newRelic` are the values the entity must have in the `instrumentation.provider` tag.
+The first tag value that matches with the entity will be the one used to build the queries.
+
+Is also important to note that the semantics of the queries should match between each implementation. This includes things like average vs counts, units and other details.
+
+If no rule matches the `newRelic` one will be used.
```
IreneMLC, line 76:
```suggestion
If no rule matches the telemetry, the `newRelic` one will be used.
```

docs/golden_metrics.md
```diff
@@ -47,7 +48,29 @@ memoryUsage:
...
 | from |   No    | `Metric` | Choose where your metric gathers the information from |
 | where |   No    | empty string | In the case you need a more granular WHERE clause added into the query you can use this field. e.g. `provider='Alb'` |
-| facet |   No    | `entity.name` | Which field to use when doing a FACET. This field will be used when displaying the metric for several entities |
+| facet |   No    | empty string | An extra facet by a specific field to be added to the default facet by entityName |
 | eventId |   No    | `entity.guid` | The event attribute used to filter the entity. We recommend to use the default `entity.guid` which is generated automatically as part of the entity synthesis. |
+| eventName |   No    | `entity.name` | The name of the field in the event that references the entity name. By default `entity.guid` which is generated automatically as part of the entity synthesis. |
 
+### Roll-up entities
 
+In the cases that the entity type can be ingested from different sources and you need to provide a different query implementation you can use `queries` instead of `query`
```
IreneMLC, line 57:
```suggestion
In the cases that the entity type can be ingested from different sources and you need to provide a different query implementation for each, you can use `queries` instead of `query`
```

IreneMLC, line 57:
```suggestion
In the cases that the entity type can be ingested from different sources and you need to provide a different query implementation for each of them, you can use `queries` instead of `query`
```

IreneMLC (approved):

docs/golden_metrics.md
naxhh: hmm this is not really matching against telemetry but the tags the entity has.

IreneMLC: True, can we change it to the tags then? (I just felt that the phrase felt a bit weird when I read it)

naxhh:
We will need a few changes for the dashboards format. So more changes will be pushed before merging. Also I'll take care of improving the docs

README.md
```diff
@@ -132,11 +132,23 @@ You can create a dashboard with the NewRelic interface and export it to JSON for
...
 ```yaml
-dashboardTemplates: 
- - ./<dashboardName>.json
+dashboardTemplates:
+ newRelic:
+  template: dashboard.json
+```
+
+If your entity has more than one telemetry source you can also provide a different dashboard for each of them. 
+The entity must have the tag `instrumentation.provider` with the value used as the key of the `dashboardTemplates`, in the example below we define a dashboard for two telemetry sources `newRelic` and `prometheus`.
```
srvaroa, line 141: Maybe worth adding which ones should people use (which at the moment will be "newRelic" if using a NR agent, or "ask us" otherwise.  We are still pending a list of sources).

IreneMLC: That makes sense, I'll add that and change the example from `prometheus` to `otherSource` to make it generic/ 

Btw, do you think that it's ok to always require people to provide at least the newRelic one? (we're also having that discussion in the Emitter because if we don't have this as the default then we'd take the first one of the list regardless of tag).

srvaroa: No, I think it's ok if NewRelic is also optional (e.g. you may have a type that is not supported by NR, but we can extract it from another integration).

apoloa (approved):

IreneMLC (approved):

PR #113, opened 2021-03-12
Add entityTagName configuration to tags

IreneMLC:
### Relevant information

Add entityTagName configuration to tags so that users can specify their desired tag key. If not specified we use the default (the attribute name)

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,15 +176,15 @@ If you want to override all the values with the new one you can configure `multi
 
 An example of this is `INFRA-CONTAINER` where the tag `container.state` will always display the last state (`running`, `stopped`, etc) instead of a list of all the states the entity has gone through.
 
-You can also configure the key-name that you want for the tags we synthesize from your telemetry attributes. 
-By default, the tag-keys are the names of the attributes you specify on the tags section, but you can specify a different key-name using the `writeAs` field. 
+You can also configure the tag name that you want for the entity tags we synthesize from your telemetry attributes. 
+By default, the entity tag names are equal to the attributes you specify on the `tags` section, but you can specify a different tag name using the `entityTagName` field. 
 
 In the example below if `attributeNameB` is present on the telemetry, a tag with key `preferredTagName` and the value of the attribute will be added to the entity. For `attributeNameC` a tag would be created with `attributeNameC` as the key. 
 
 ```yaml
   tags:
     attributeNameB:
-      writeAs: preferredTagName,
+      entityTagName: preferredTagName,
     attributeNameC:
 ```
 
--- a/README.md
+++ b/README.md
@@ -176,18 +176,20 @@ If you want to override all the values with the new one you can configure `multi
 
 An example of this is `INFRA-CONTAINER` where the tag `container.state` will always display the last state (`running`, `stopped`, etc) instead of a list of all the states the entity has gone through.
 
-You can also configure the tag name that you want for the entity tags we synthesize from your telemetry attributes. 
-By default, the entity tag names are equal to the attributes you specify on the `tags` section, but you can specify a different tag name using the `entityTagName` field. 
+You can also configure the tag name that you want for the entity tags we synthesize. By default, the entity tag names are equal to the telemetry attributes you specify on the `tags` section, but you can specify a different tag name using the `entityTagName` field.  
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -176,6 +176,18 @@ If you want to override all the values with the new one you can configure `multi
 
 An example of this is `INFRA-CONTAINER` where the tag `container.state` will always display the last state (`running`, `stopped`, etc) instead of a list of all the states the entity has gone through.
 
+You can also configure the key-name that you want for the tags we synthesize from your telemetry attributes.
```
srvaroa, line 179: `s/tags/entity tags` (just to make it extra explicit as the fact that we have two different sets of tags/attributes can be confusing for people)

IreneMLC: Names sense, will change, thanks!

README.md
```diff
@@ -176,6 +176,18 @@ If you want to override all the values with the new one you can configure `multi
...
 
+You can also configure the key-name that you want for the tags we synthesize from your telemetry attributes. 
+By default, the tag-keys are the names of the attributes you specify on the tags section, but you can specify a different key-name using the `writeAs` field. 
+
+In the example below if `attributeNameB` is present on the telemetry, a tag with key `preferredTagName` and the value of the attribute will be added to the entity. For `attributeNameC` a tag would be created with `attributeNameC` as the key. 
+
+```yaml
+  tags:
+    attributeNameB:
+      writeAs: preferredTagName,
```
srvaroa, line 187: (I know I proposed rewriteAs sorry, my brain works slowly)

Would `entityTag` be easier to understand?

definitions/infra-container/definition.yml
```diff
@@ -39,21 +42,36 @@ synthesis:
...
+    instrumentation.provider:
+    instrumentation.name:
+    # For OpenTelemetry
+    telemetry.sdk.name:
+      writeAs: instrumentation.provider
+    telemetry.sdk.language:
+      writeAs: language
+    # AWS tags
+    aws.region:
+    aws.ec2.instanceType:
```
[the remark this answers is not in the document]
srvaroa: aws.* yes, instrumentation.provider + name no, they are added implicitly by the synthesis service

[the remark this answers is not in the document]
IreneMLC: I checked the existing container entities that we have in our NR accounts and they don't seem to have these tags btw, but I guess that customers with different telemetry sources might

README.md
IreneMLC: How about `entityTagName`? it might indeed be clearer that re/writeAs

srvaroa: Like.

README.md
```diff
@@ -176,6 +176,18 @@ If you want to override all the values with the new one you can configure `multi
 
 An example of this is `INFRA-CONTAINER` where the tag `container.state` will always display the last state (`running`, `stopped`, etc) instead of a list of all the states the entity has gone through.
 
+You can also configure the tag name that you want for the entity tags we synthesize from your telemetry attributes.
```
srvaroa, line 179: The first sentence seems mostly redundant with the second. Would it make sense to remove it?

README.md
```diff
@@ -176,6 +176,18 @@ If you want to override all the values with the new one you can configure `multi
...
 
+You can also configure the tag name that you want for the entity tags we synthesize from your telemetry attributes. 
+By default, the entity tag names are equal to the attributes you specify on the `tags` section, but you can specify a different tag name using the `entityTagName` field. 
+
+In the example below if `attributeNameB` is present on the telemetry, a tag with key `preferredTagName` and the value of the attribute will be added to the entity. For `attributeNameC` a tag would be created with `attributeNameC` as the key. 
+
+```yaml
+  tags:
+    attributeNameB:
+      entityTagName: preferredTagName,
```
srvaroa, line 187: Is the trailing comma correct?

README.md
```diff
@@ -176,6 +176,18 @@ If you want to override all the values with the new one you can configure `multi
...
+    attributeNameC:
+```
+
```
srvaroa, line 190: I would add here a recommendation on when to use it.  E.g. not arbitrarily copy values, but use it to create consistency in entities that get data from various sources, referencing the container case for example.

srvaroa (commented):
LGTM, just minor comments.  Thanks!

README.md
IreneMLC: Ups no, thanks!

README.md
IreneMLC: True, I'll reword it

srvaroa (approved):

PR #114, opened 2021-03-15
adding ext-router entity definition and metrics

thezackm:
### Relevant information

adding `ROUTER` entity for the `EXT` domain. Original definition is based on network device metrics sent via Kentik technical partnership, but this definition is being kept ambiguous to allow for other technical partnerships to use the entity in the future.

`ROUTER` relates to any L3 device in the networking space that has the capability to route packets based on defined protocols. This entity can also be decorated with Flow telemetry (NetFlow/sFlow/jFlow/IPFIX/etc), among other things. 

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-router/definition.yml
+++ b/definitions/ext-router/definition.yml
@@ -8,10 +8,8 @@ synthesis:
   encodeIdentifierInGUID: true
 
   conditions:
-    - attribute: metricName
-      prefix: kentik
     - attribute: provider
-      value: router
+      value: kentik-router
 
   tags:
     src_addr:
--- a/definitions/ext-router/definition.yml
+++ b/definitions/ext-router/definition.yml
@@ -1,0 +1,26 @@
+domain: EXT
+# Network Router devices from Kentik using Metric API
+# Leaving this ambiguous to allow future technical partners to use the same entity type with additional conditions
+type: ROUTER
+synthesis:
+  name: device_name
+  identifier: device_name
+  encodeIdentifierInGUID: true
+
+  conditions:
+    - attribute: metricName
+      prefix: kentik
+    - attribute: provider
+      value: router
+
+  tags:
+    src_addr:
+
+compositeMetrics:
+  goldenMetric:
+    - golden_metrics.yml
+  summaryMetrics:
+    - summary_metrics.yml
+
+goldenTags:
+  - src_addr
--- a/definitions/ext-router/definition.yml
+++ b/definitions/ext-router/definition.yml
@@ -8,17 +8,17 @@ synthesis:
   encodeIdentifierInGUID: true
 
   conditions:
-    - attribute: provider
-      value: kentik-router
+  - attribute: provider
+    value: kentik-router
 
   tags:
     src_addr:
 
 compositeMetrics:
-  goldenMetric:
-    - golden_metrics.yml
+  goldenMetrics:
+  - golden_metrics.yml
   summaryMetrics:
-    - summary_metrics.yml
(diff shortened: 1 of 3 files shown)


definitions/ext-router/definition.yml
```diff
@@ -0,0 +1,26 @@
+domain: EXT
+# Network Router devices from Kentik using Metric API
+# Leaving this ambiguous to allow future technical partners to use the same entity type with additional conditions
+type: ROUTER
+synthesis:
+  name: device_name
+  identifier: device_name
+  encodeIdentifierInGUID: true
+
+  conditions:
```
IreneMLC, line 10: we'll only really apply 1 of these conditions atm (the telemetry won't have to match both), right @srvaroa?

thezackm: this telemetry's current design will require matching both; if we can't do that please let me know ASAP so I can go back to the vendor and request a change. thanks!!!

definitions/ext-router/golden_metrics.yml
```diff
@@ -0,0 +1,27 @@
+cpuUtilization:
+  title: CPU Utilization (%)
+  query:
+    select: average(kentik.snmp.CPU) AS 'CPU Utilization'
+    from: Metric
+    where: "provider = 'kentik-router'"
```
srvaroa, line 6: Same as in the other PR, I think this where is not necessary (pending for @naxhh  to confirm)

srvaroa (commented):
LGTM! thanks for adding these (I'll wait for @naxhh's reply on a minor comment before merging)

definitions/ext-router/definition.yml
IreneMLC: Thaks for the change!

IreneMLC (approved):

definitions/ext-router/summary_metrics.yml
```diff
@@ -0,0 +1,41 @@
...
+  title: Router IP Address
+  unit: STRING
+  tag:
+    key: src_addr
+
+cpuUtilization:
+  title: CPU Utilization (%)
+  unit: PERCENTAGE
+  query:
+    select: average(kentik.snmp.CPU) AS 'CPU Utilization'
```
naxhh, line 11: The `as` here will not be really used anywhere. And I'm not sure how our summary metrics will handle that.
@gvassallo would appreciate your input on this one.

naxhh (commented):
Aside of the comment LGTM

definitions/ext-router/summary_metrics.yml
gvassallo: i think this might break summary metrics or in general, it's not required

thezackm: @gvassallo | @naxhh - so I should remove all of the `'AS'` on these PRs for summary metrics only, or both summary and golden?

gvassallo: I'd say so, we're gonna add the `as` to the golden metric queries by default in one of our next releases. I'd leave @naxhh making the final decision though :D

naxhh: @thezackm remove them all from both sides. We are in the process of automating this and will be easier for us.
Thanks for the patience!


Side note: I'll make a note to check SM behavior on this so we can add it to the linting.

thezackm: done and done. thanks!

naxhh (approved):

PR #115, opened 2021-03-15
adding ext-switch entity definition and metrics

thezackm:
### Relevant information

adding `SWITCH` entity for the `EXT` domain. Original definition is based on network device metrics sent via Kentik technical partnership, but this definition is being kept ambiguous to allow for other technical partnerships to use the entity in the future.

`SWITCH` relates to any L2 device in the networking space that lacks the capability to route packets.

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-switch/golden_metrics.yml
+++ b/definitions/ext-switch/golden_metrics.yml
@@ -1,0 +1,27 @@
+cpuUtilization:
+  title: CPU Utilization (%)
+  query:
+    select: average(kentik.snmp.CPU) AS 'CPU Utilization'
+    from: Metric
+    where: "provider = 'switch'"
+
+memoryUtilization:
+  title: Memory Utilization (%)
+  query:
+    select: average(kentik.snmp.MemoryUtilization) AS 'Memory Utilization'
+    from: Metric
+    where: "provider = 'switch'"
+
+receiveErrors:
+  title: Aggregate Receive Errors (count)
+  query:
+    select: sum(kentik.snmp.ifInErrors) AS 'Receive Errors'
+    from: Metric
+    where: "provider = 'switch'"
+
+transmitErrors:
+  title: Aggregate Transmit Errors (count)
+  query:
+    select: sum(kentik.snmp.ifOutErrors) AS 'Transmit Errors'
+    from: Metric
+    where: "provider = 'switch'"
--- a/definitions/ext-switch/golden_metrics.yml
+++ b/definitions/ext-switch/golden_metrics.yml
@@ -3,25 +3,25 @@ cpuUtilization:
   query:
     select: average(kentik.snmp.CPU) AS 'CPU Utilization'
     from: Metric
-    where: "provider = 'switch'"
+    where: "provider = 'kentik-switch'"
 
 memoryUtilization:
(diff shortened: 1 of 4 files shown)


definitions/ext-switch/golden_metrics.yml
```diff
@@ -0,0 +1,27 @@
+cpuUtilization:
+  title: CPU Utilization (%)
+  query:
+    select: average(kentik.snmp.CPU) AS 'CPU Utilization'
+    from: Metric
+    where: "provider = 'kentik-switch'"
```
srvaroa, line 6: The `where` clauses should not be necessary I think as the GMs will match against telemetry that is already stamped with a GUID.  Is that right @naxhh ?

srvaroa (commented):
LGTM, just a minor question pending for @naxhh 's feedback.

definitions/ext-switch/golden_metrics.yml
IreneMLC: I think that for that to happen you'd have to use [queries](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/docs/summary_metrics.md#roll-up-entities), it might make sense here:

> In the cases that the entity type can be ingested from different sources and you need to provide a different query implementation you can use queries instead of query

thezackm: for context; this is using the same pattern as multiple [AWS Entities](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/definitions/infra-awsalb/golden_metrics.yml) in that `provider` can have multiple values so the `WHERE` seems helpful.

If it's not required because there's an implicit `WHERE entityGuid = '1234abcd'`; then I would agree it's unnecessary.

srvaroa: Yeah, when the queries get executed they will be applied to a specific entity, so the `where` would be redundant. But it doesn't break anything anyway.

thezackm: ok; would you like me to remove them @srvaroa? I can get that done real quick so this can get merged.

thezackm: also; would the same apply to the Summary Metrics?

srvaroa: Yes! I'll ping @naxhh  to get this approved soon as possible on our end.

srvaroa: @thezackm let's leave it for now, we're not 100% sure if that where is important for golden signals.  I can clean it up later if needed.

naxhh: As talking with @srvaroa I have a doubt on a specific use case. So let's keep the provider for now and I'll double check that.

naxhh (approved):
LGTM.

We need to double check some behaviors in our side before merging

srvaroa (approved):

srvaroa:
@thezackm have you signed the CLA? (I have seen the bot failed to report status on several PRs so it may be that it just broke).

thezackm:
@srvaroa  yes I have. I agree it might be broken, this happened to me on another repo recently as well.

PR #117, opened 2021-03-17
Add a git repo definition

jerelmiller:
### Relevant information

Add a git repo entity definition to allow the Entity Community team to execute on the steel thread.

### Checklist

 explanation above.

### DIFF
--- a/definitions/res-gitrepository/definition.yml
+++ b/definitions/res-gitrepository/definition.yml
@@ -1,0 +1,5 @@
+domain: RES
+type: GITREPOSITORY
+configuration:
+  entityExpirationTime: QUARTERLY
+  alertable: false


srvaroa (approved):
Thanks!

cianBuckley (approved):

definitions/git-repository/definition.yml
```diff
@@ -0,0 +1,5 @@
+domain: GIT
```
srvaroa, line 1: Hi, we're pending to confirm if this is the domain we want to use with @jfjoly .

We'll likely prefer to use RES- (for resource, or entities that are not obviously linked to the ordinary ones that emit telemetry) or EXT- here (for the rest).

The rationale is that the domain field exists for legacy reasons so we'd prefer to avoid putting much semantics in the domain.  We should be able to clear this during today.

srvaroa (changes requested):
Marking request changes just to prevent accidental merges while we confirm the best domain to use, but LGTM.

Thanks!

definitions/git-repository/definition.yml
jerelmiller: Good to know! I will make this change!

srvaroa (approved):

srvaroa:
It looks like the CLA check got lost. @jerelmiller can you re-push this so it triggers? (I can't do it since it's on your fork)

srvaroa (approved):

srvaroa:
Merging this one, I asked @jerelmiller if we want to have synthesis rules also, but they can be added in another PR

definitions/res-gitrepository/definition.yml
```diff
@@ -0,0 +1,5 @@
+domain: RES
+type: GITREPOSITORY
```
naxhh, line 2: nit: Should we GIT_REPOSITORY instead?

naxhh (approved):

definitions/res-gitrepository/definition.yml
srvaroa: I don't have a strong opinion, these are not meant to be user visible anyway

jerelmiller: I only went with this approach since it looked like a lot of the AWS related types did the same (i.e. `AWSIAMOPENIDPROVIDER`).Totally happy to change this if we like the `GIT_REPOSITORY` instead.

jerelmiller:
@srvaroa I signed the CLA, and just did it again, but not sure why the status is not changing. You can clearly see the comment above is green. Let me reach out to our OSS team.

definitions/res-gitrepository/definition.yml
naxhh: whatever you prefer :D the names where without the `_` because the limit was too low before. I'm ok with any variation.

naxhh:
@jerelmiller the CLA bot seems to be broken. We will need to ping other teams to look into that one

PR #119, opened 2021-03-18
Added SNMP Lenovo Compute Fans definition

harrykimpel:
### Relevant information

Added SNMP Lenovo Compute Fans definition

### Checklist


definitions/infra-snmplenovocomputefans/definition.yml
```diff
@@ -0,0 +1,9 @@
+domain: INFRA
```
srvaroa, line 1: Hi @harrykimpel , we're preferring to use EXT here as a general rule.  Can you update this value here, and the `infra-` prefix in the folder containing these files?

srvaroa (commented):
LGTM, just a minor correction.

thanks for contributing it!

definitions/infra-snmplenovocomputefans/definition.yml
harrykimpel: alright, makes sense ... will update

harrykimpel: by the way: are Events already supported at all?

definitions/ext-snmpdevice/definition.yml
```diff
@@ -0,0 +1,9 @@
+domain: EXT
+type: SNMPDEVICE
+goldenTags: []
+compositeMetrics:
+  goldenMetrics:
+  - golden_metrics.yml
+configuration:
```
srvaroa, line 7: One extra question: since the definitions don't add synthesis rules these entities will only appear if they are indexed directly.  Was that your intent (it's fine if you want to register the type), or were you expecting the entities to be synthesized out of telemetry data?

naxhh (commented):
Could you explain a bit more on this type?

I'm wondering if we should have a generic "fan" type or this is something more than that.

Also as galo commented i don't see a synthesis section so not sure how this will be ingested.

definitions/infra-snmplenovocomputefans/definition.yml
harrykimpel: @srvaroa

definitions/ext-snmpdevice/definition.yml
harrykimpel: just added a synthesis rule, hope this makes sense

harrykimpel:
> Could you explain a bit more on this type?
> 
> I'm wondering if we should have a generic "fan" type or this is something more than that.
> 
> Also as galo commented i don't see a synthesis section so not sure how this will be ingested.

@naxhh I am adding some SNMP data via the https://github.com/newrelic-experimental/nri-snmp-trapd integration. One of the traps I have defined will give me the health of the fan(s) of a device. there are other attributes as well on the same event. And, there are also other events generated with a different event name.

definitions/ext-snmpdevice/definition.yml
IreneMLC: Since just the name of the identifier attribute is usually too generic to identify an entity we also require that you add a condition (that's why the build is failing), maybe the event name should be the condition here?

harrykimpel: @IreneMLC like so?

srvaroa: Right now the synthesis engine only supports a single condition. Would name = fanTable be enough to match the right telemetry for this case?

IreneMLC: Hi Harry, the conditions have [this format](https://github.com/newrelic-experimental/entity-synthesis-definitions#conditions) in your case it could be something like this, but could you let us know if there's an account that we can impersonate to check this telemetry and be able to give you better guidance? thanks!

```
conditions:
    # The attribute’s value must match the provided value
    - attribute: name
      value: eventName
```

naxhh:
@harrykimpel The point I'm trying to make is that an entity models an object.

We have hosts, applications and pods. We don't really care if they are ingested via NR agent, prometheus or via SNMP.
We always have a "tangible" entity.

In this case it seems that we are modeling a FAN. In this specific case that fan sends health status via SNMP.
If in the future someone makes an integration to send fan data we would probably aggregate that here.
Would that statement be correct?

An SNMPDEVICE seems to vague. Will be that a router? a switch? a fan? all of them? It seems that we are doing a bucket of "things that report using SNMP".
But again. I lack total context on this so I'm probably wrong on this one.

srvaroa:
I'm checking the telemetry for these devices as we have noticed that they already bring a GUID in some cases so that'll make them skip the synthesis process. 

I'll also make some rounds gathering opinions on whether SNMPDEVICE is too broad (it can indeed contain many different "things") or we're ok with that for now.

@harrykimpel I may take a couple of days since I have to get in touch with several people but we'll report back here asap and get this working.

harrykimpel:
> I'm checking the telemetry for these devices as we have noticed that they already bring a GUID in some cases so that'll make them skip the synthesis process.
> 
> I'll also make some rounds gathering opinions on whether SNMPDEVICE is too broad (it can indeed contain many different "things") or we're ok with that for now.
> 
> @harrykimpel I may take a couple of days since I have to get in touch with several people but we'll report back here asap and get this working.

@srvaroa did you have a chance to look at this? any suggestion on how to move this forward?

srvaroa:
Hi @harrykimpel, I discussed with the maintainers in the agent and the best path forward would be to add a configuration parameter in the agent that turns off sending the `entity.guid` attribute in the telemetry.  That'll allow the synthesis engine to inspect it and generate entities.

harrykimpel:
> Hi @harrykimpel, I discussed with the maintainers in the agent and the best path forward would be to add a configuration parameter in the agent that turns off sending the `entity.guid` attribute in the telemetry. That'll allow the synthesis engine to inspect it and generate entities.

Hi @srvaroa , so this is not something I can change, right? agent team would need to get rid of the entityGuid in the actual event of the integration (SNMPLenovoComputeFans in this case), right?

srvaroa:
@harrykimpel exactly!

entityBot (changes requested):
beep boop bop.

 ```
* Error VALIDATING domainType: 'INFRA-VARNISHINSTANCE' on summaryMetrics for key 'threads newRelic' - DiracParserException: Error at line 1 position 16, perhaps unterminated `

* Error VALIDATING domainType: 'INFRA-AWSKINESISDELIVERYSTREAM' on summaryMetrics for key 'delivered newRelic' - DiracParserException: Error at line 1 position 152, unexpected 'second'
```

entityBot (approved):

entityBot (approved):

entityBot (approved):

naxhh:
@srvaroa could you TLDR what is missing here? I understand the NR agent requires some changes?

entityBot (approved):

srvaroa:
@naxhh yeah, it looks like this telemetry already brings a GUID, so the rule will have no effect.  We discussed above about the agent removing the guid so that the rule engine applies, but from what I know this was not entirely feasible.

(rebased the branch to keep it updated)

definitions/ext-snmpdevice/definition.yml
srvaroa: @harrykimpel This entity (snmpdevice) does not include any synthesis rules, was that intentional? Is there a plan to index them in any other way?

cianBuckley:
Closing due to inactivity

PR #120, opened 2021-03-19
CBO-188 Gateway - Entity Synthesis for MySQL

AustinNester:
Definitions for metrics related to Datadog's MySQL implementation (https://docs.datadoghq.com/integrations/mysql/?tab=host#data-collected)
Currently only golden metrics have been defined, but there is a possibility that some of the metrics will be summary metrics.  I will add a comment to those that I've seen

### Relevant information

Definitions for metrics related to Datadog's MySQL implementation (https://docs.datadoghq.com/integrations/mysql/?tab=host#data-collected)
Currently only golden metrics have been defined, but there is a possibility that some of the metrics will be summary metrics,  I will add a comment to those that I've seen.  Tags may also come later as we move forward. Thank you!

### Checklist

 explanation above.


definitions/ext-ddmysql/golden_metrics.yml
```diff
@@ -0,0 +1,1462 @@
...
+    eventName: entity.name
+wsrepFlowControlSent:
+  title: Galera Node Sent Pausing Flow Control Message
+  query:
+    select: mysql.galera.wsrep_flow_control_sent
+    from: Metric
+    eventId: entity.guid
+    eventName: entity.name
+wsrepCertDepsDistance:
+  title: Average Distance Between Lowest and Highest SeqNo Node Can Apply in Parallel
```
AustinNester, line 610: This is potentially a summary metric, but I wasn't sure since it is technically already averaged by the time it is given to us. (I'm new to all of the terminology and such so please correct me if I am misunderstanding anything :) )

definitions/ext-ddmysql/definition.yml
```diff
@@ -0,0 +1,18 @@
+domain: EXT
+type: DATADOG_MYSQL
```
cianBuckley, line 2: Could you match the directory name with this name provided here?

definitions/ext-ddmysql/golden_metrics.yml
cianBuckley: The difference will be in how they are displayed, being that summary metrics go to a table.

definitions/ext-ddmysql/golden_metrics.yml
```diff
@@ -0,0 +1,1462 @@
+bufferPoolDirty:
```
cianBuckley, line 1: There are a lot of metrics here.. i would feel too many? How do we internally handle GM limits @IreneMLC ?

definitions/ext-ddmysql/definition.yml
```diff
@@ -0,0 +1,18 @@
+domain: EXT
```
cianBuckley, line 1: i think EXT is fine, right @IreneMLC ?

definitions/ext-ddmysql/golden_metrics.yml
IreneMLC: Yes, we have a limit of 10 GMs per entity type ([docs](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/docs/golden_metrics.md#golden-metrics)) but it's not enforced on the repo's check, I'll look into adding it.

cianBuckley: @austinNester You could create a dashboard with all of these metrics. Ideally, we should only have a small handful of golden metrics (favourite metrics) approx 4 should be ideal. as Irene says, 10 is the limit

definitions/ext-ddmysql/definition.yml
```diff
@@ -0,0 +1,18 @@
...
+type: DATADOG_MYSQL
+
+synthesis:
+  name: dd.mysql
+  identifier: dd.mysql
+  encodeIdentifierInGUID: false
```
naxhh, line 7: If `dd.mysql` values are the equivalents of a hostname this should be true

definitions/ext-ddmysql/golden_metrics.yml
```diff
@@ -0,0 +1,1462 @@
...
+  title: Buffer Pool Dirty (byte)
+  displayAsValue: false
+  query:
+    select: mysql.innodb.buffer_pool_dirty
+    from: Metric
```
naxhh, line 6: Golden Signals has no support for shimming metrics yet.
So there are two possibilities here:
- keep these queries and don't have golden signals (lookout view) until they implement support for it.
- don't use the shimming and use the underlying queries and have golden signals

definitions/ext-ddmysql/golden_metrics.yml
naxhh: So to make a recap:

Summary metrics are displayed in the explorer table.
Golden metrics can be displayed in a bunch of places. And we normally display up to 4 (but 10 is the real limit)

If you want to display what data we have for an entity in the entity view you would create a dashboard for that.

definitions/ext-ddmysql/definition.yml
AustinNester: I ended up realizing I didn't actually understand what these are derived from, so it looks like the value will actually be `host.name` based on what we got from the metric.  Switching the boolean to true as well.

definitions/ext-ddmysql/golden_metrics.yml
AustinNester: Ah I see, thank you both!

definitions/ext-ddmysql/golden_metrics.yml
AustinNester: What are "shimming metrics" if I may ask?

AustinNester:
Thank you all so much for your patience and feedback! Apologies for all of my confusion, I'm super new and didn't fully understand how metrics or telemetry in general work under the hood.  With the feedback I went and dug through some other entity definitions and compared them to the metrics in NR, so I think I (hopefully) understand now! I'll need to grab some additional information from my team, and maybe make some decisions about changes to the data we are sending in before finishing this up.  Should I close the PR for now, or just update it when I get further information?

definitions/ext-datadog_mysql/definition.yml
```diff
@@ -0,0 +1,18 @@
+domain: EXT
+type: DATADOG_MYSQL
```
srvaroa, line 2: We might want to make this a more generic EXT-MYSQL.  Let me check with product management on our side.

srvaroa: Ok I have confirmed with @jfjoly that we should make this more generic, can you change the type to `MYSQL`?

Also, you'll have to update the directory to be `ext-mysql`.

Thanks!

srvaroa (changes requested):
Just the name of the topic, otherwise looks ok for me

definitions/ext-ddmysql/golden_metrics.yml
naxhh: Shimmings is something NRDB team is working on. But basically the objective is to always use `FROM Metric` in the queries.
So it converts "old" ones like `FROM AwsEc2Sample` into a query that works using `FORM Metric`. This is invisible for users but sadly we have a service that "knows" about this things. And hence we should too :(

That said that service should implement this logic and we should not really care about it.

definitions/ext-datadog_mysql/definition.yml
AustinNester: Switched over to MYSQL and `ext-mysql` for the directory!

definitions/ext-datadog_system/definition.yml
```diff
@@ -0,0 +1,18 @@
+domain: EXT
+type: DATADOG_SYSTEM
```
srvaroa, line 2: I thought this was meant to be MYSQL? (`type` and the directory)

srvaroa: Oh ignore this, I didn't realize it's a separate entity

naxhh: i'm wondering if system is a more broad definition that could include other types than host? That said +1 to using EXT-HOST

definitions/ext-datadog_system/golden_metrics.yml
```diff
@@ -0,0 +1,19 @@
+systemLoad:
+  title: System Load
+  query:
+    select: sum(system.load.1), sum(system.load.5), sum(system.load.15)
```
naxhh, line 4: remember that golden metrics are the most important metrics to look at.
In this case which of the 3 is the most important one? (maybe load.15 or load.5?)

I would say the same for the rest.
CPU we normally look into usage not system, iowait, etc...

naxhh (commented):
LGTM i would keep the EXT-DD-system golden metrics to just 1 value to display instead of several of them.

definitions/ext-datadog_system/golden_metrics.yml
AustinNester: The golden metrics here were given as a whole by the authors of the ticket, so I will relay the feedback.  I will also see if they would be ok changing the multiple value metrics into dashboards instead of golden metrics.

definitions/ext-datadog_system/definition.yml
AustinNester: A question on this: There is already an `ext-host` directory and definition, what should be the approach for changing this over to  `ext-host`?

entityBot (approved):

definitions/ext-datadog_system/definition.yml
AustinNester: Bumping this question and giving a bit more background: these entities are to support specifically Datadog entries coming in from an internal service hooked up to a datadog agent, with a datadog. prefix we are appending upon intake.  Should they still be switched over to `EXT-HOST` and `EXT-MYSQL` in that instance, or wouldn't that make them too generic?

definitions/ext-datadog_system/definition.yml
```diff
@@ -0,0 +1,18 @@
...
+
+synthesis:
```
cianBuckley, line 4: This synthesis rule can go in this file here:
https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/definitions/ext-host/definition.yml

ext-host's current definition has no synthesis rules, so yours will be the first! Same goes for the golden metrics and summary metrics for ex-host - there are currently none, so you can just go ahead and put all of yours in there. 

For an example of an entity type with multiple rules (from multiple providers), you can look at https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/definitions/ext-service/definition.yml#L4 
This has 3 rules, executed in order

cianBuckley: Provided you change over directory of the metrics and add the synthesis rule to ext-host, i think we're pretty good to merge!

entityBot (approved):

entityBot (approved):

definitions/ext-datadog_system/definition.yml
AustinNester: Awesome, moved it over to ext-host. For the WIP stuff to add multiple providers, should I throw a ticket into our backlog to come back and update the metrics when it is released?  I went ahead and created tickets on our side to add some tags so we can be ready to implement it. 

Was also going to move some of the metrics over to dashboards as @naxhh suggested, but that will also need to wait until we get some tags implemented on our side in the above mentioned ticket.

entityBot (approved):

definitions/ext-datadog_system/definition.yml
cianBuckley: @AustinNester we will be able to determine source of telemetry via the `instrumentation.provider=datadog` tag that will be pinned to all data points

definitions/ext-datadog_system/definition.yml
cianBuckley: @AustinNester considering datadog is the first use case for these two entity types, you'll not need to wait for multiple sources - the next PR will!

Moving some of the metrics would make most sense to do it now. Can you get this done today? I can merge your PR once you're done

cianBuckley:
Looking good! You've mentioned you want to move some of the metrics out and into a dashboard, I'll hold off on merging until you do that but apart from that all looks good to me. 
Thanks!

srvaroa (approved):
thanks!

entityBot (approved):

definitions/ext-datadog_system/definition.yml
AustinNester: @cianBuckley Sure thing, moved the metrics over to dashboard.json, let me know if it looks correct!

cianBuckley (approved):

entityBot (approved):

naxhh (approved):

naxhh:
Seems the validator complains because this PR is using the old dashboards format.

```
../definitions/ext-host/dashboard.json invalid
[
  {
    "keyword": "additionalProperties",
    "dataPath": "",
    "schemaPath": "#/additionalProperties",
    "params": {
      "additionalProperty": "title"
    },
    "message": "should NOT have additional properties"
  }
]
```

That can be fixed importing that format into a dashboard and exporting again. So it produces the new format

naxhh:
Also @AustinNester would be great if you can sign the CLA

entityBot (approved):

AustinNester:
> Also @AustinNester would be great if you can sign the CLA
Looks like it is waiting on @cianBuckley to sign the CLA as well

entityBot (approved):

entityBot (approved):

srvaroa (approved):

PR #123, opened 2021-03-22
Add definition and synthesis rule for KUBERNETES_POD

douglascamata:
### Relevant information

This adds support for indexing Kubernetes Pods as entities. 

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-kubernetespod/definition.yml
+++ b/definitions/infra-kubernetespod/definition.yml
@@ -1,0 +1,35 @@
+domain: INFRA
+type: KUBERNETES_POD
+configuration:
+  entityExpirationTime: DAILY
+  alertable: true
+
+synthesis:
+  rules:
+  - identifier: entity.id
+    name: k8s.podName
+    encodeIdentifierInGUID: true
+    conditions:
+    - attribute: k8s.pod.isReady
+    tags:
+      k8s.status:
+      k8s.clusterName:
+      k8s.namespaceName:
+      k8s.podName:
+      k8s.nodeName:
+      k8s.pod.createdAt:
+      k8s.pod.startTime:
+      k8s.pod.isReady:
+      k8s.pod.isScheduled:
+      label.topology.kubernetes.io/region:
+      label.topology.kubernetes.io/zone:
+      label.eks.amazonaws.com/compute-type:
+      label.kubernetes.io/arch:
+      label.kubernetes.io/hostname:
+      label.kubernetes.io/os:
+  tags:
+    newrelic.integrationName:
+    newrelic.integrationVersion:
+    newrelic.agentVersion:
+    aws.region:
+    aws.ec2.instanceType:
--- a/definitions/infra-kubernetespod/definition.yml
+++ b/definitions/infra-kubernetespod/definition.yml
@@ -5,32 +5,30 @@ configuration:
   alertable: true
 
 synthesis:
-  rules:
-  - identifier: entity.id
-    name: k8s.podName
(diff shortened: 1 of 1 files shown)


definitions/infra-kubernetespod/definition.yml
```diff
@@ -0,0 +1,35 @@
...
+    tags:
+      k8s.status:
+      k8s.clusterName:
+      k8s.namespaceName:
+      k8s.podName:
+      k8s.nodeName:
+      k8s.pod.createdAt:
+      k8s.pod.startTime:
+      k8s.pod.isReady:
+      k8s.pod.isScheduled:
```
naxhh, line 23: Shouldn't these two be `multiValue: false` ? Same probably with status

naxhh (commented):
LGTM aside of the multivalue

definitions/infra-kubernetespod/definition.yml
douglascamata: I don't know. In the definition of the INFRA Container entity there's no multivalue. Maybe it's missing something there?

IreneMLC: [Here's](https://github.com/newrelic-experimental/entity-synthesis-definitions#tags) an example btw

naxhh: @douglascamata yep we are probably missing them there too. Since the multiValue support was added not long ago here.

IreneMLC (approved):

definitions/infra-kubernetespod/definition.yml
```diff
@@ -0,0 +1,38 @@
...
+  alertable: true
+
+synthesis:
+  identifier: entity.id
+  name: k8s.podName
+  encodeIdentifierInGUID: true
+  conditions:
+  - attribute: k8s.pod.isReady
+  tags:
+    k8s.status:
```
srvaroa, line 14: Should this add aliasing? (same for `isReady` and `isScheduled` below)

srvaroa (commented):
LGTM, minor question

definitions/infra-kubernetespod/definition.yml
douglascamata: This doesn't need aliasing here. Only the k8s container will need both tags `k8s.status` and `container.state`.

douglascamata: @srvaroa is this yaml already supporting more than 1 rule per attribute so that we can achieve what I mention in the comment above?

IreneMLC: Indeed, it supports it, you can define different rules with their specific tags for a single entity, container is a good example https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/definitions/infra-container/definition.yml#L32

douglascamata: So I can do this?

```yaml
k8s.status:
        entityTagName: container.state
k8s.status:
```

And I would get the `k8s.status` and `container.state` tags?

IreneMLC: Not exactly since`k8s.status` would be a duplicated map key but if you are doing this you probably have 2 sources of telemetry and want to have two rules, so you could do something like this:

```
rules:
  - identifier: attributeA
  ....
  tags:
  k8s.status:
    entityTagName: container.state
 - identifier: attributeA
   conditions:
      .....
  ....
  tags:
  k8s.status
```

PR #125, opened 2021-03-23
Add Kubernetes deployments synthesis rules and update the pod's

douglascamata:
### Relevant information

Add rules for Kubernetes deployments. Plus update the pods' to include more tags: `k8s.deploymentName`,  `k8s.createdBy` and `k8s.createdKind`.

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-kubernetespod/definition.yml
+++ b/definitions/infra-kubernetespod/definition.yml
@@ -18,6 +18,8 @@ synthesis:
     k8s.podName:
     k8s.nodeName:
     k8s.deploymentName:
+    k8s.createdBy:
+    k8s.createdKind:
     k8s.pod.createdAt:
     k8s.pod.startTime:
     k8s.pod.isReady:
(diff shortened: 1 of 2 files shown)


definitions/infra-kubernetespod/definition.yml
```diff
@@ -17,6 +17,9 @@ synthesis:
     k8s.namespaceName:
     k8s.podName:
     k8s.nodeName:
+    k8s.deploymentName:
```
IreneMLC, line 20: is this for k8sPod or k8sDeployment?

douglascamata: For both. We also include `k8s.deploymentName` in pod metrics, in case the pods were created by a Deployment.

IreneMLC: thanks! (just curious).

IreneMLC (approved):

PR #126, opened 2021-03-23
Fix encodeIdentifierInGUID for Kubernetes pods and containers

douglascamata:
### Relevant information

Fix the `encodeIdentifierInGUID` setting for Kubernetes pods and containers.

### Checklist

 explanation above.


IreneMLC (approved):
Thanks!

PR #127, opened 2021-03-23
Fix: add provider filter for RDS

jlegoff:
### Relevant information

Added the provider filter to RDS entities. We're seeing shimming errors and we believe it might be because of this missing filter.

### Checklist

 explanation above.


IreneMLC:
This is related to the slack discussion with Jose, right @jlegoff?

jlegoff:
@IreneMLC yes!

IreneMLC (approved):
Thanks!

PR #133, opened 2021-03-25
Add k8s daemonset synthesis rule

douglascamata:
### Relevant information

We're considering to rewrite all the `k8s.<entity_name>.createdAt` tags as `k8s.createdAt`. It will be done in another PR if we decide to move forward with it. 

### Checklist

 explanation above.


definitions/infra-kubernetesdaemonset/definition.yml
```diff
@@ -0,0 +1,32 @@
+domain: INFRA
+type: KUBERNETES_DAEMONSET
```
IreneMLC, line 2: The folder name must match this

IreneMLC (approved):
Looks good!

douglascamata:
We decided to drop the `createdAt` tag altogether.

definitions/infra-kubernetes_daemonset/definition.yml
```diff
@@ -0,0 +1,31 @@
+domain: INFRA
+type: KUBERNETES_DAEMONSET
+configuration:
+  entityExpirationTime: DAILY
+  alertable: true
+
+synthesis:
+  identifier: entity.id
+  name: k8s.daemonsetName
+  encodeIdentifierInGUID: false
```
naxhh, line 10: I'm wondering if things that come from a user-generated name should be always encoded.
My thinking is that, same as hostname, we don't control lengths and it could be over the limit.

@srvaroa any input on this one?

naxhh (commented):
Shouldn't we provide a dashboard for the entity?

definitions/infra-kubernetes_daemonset/definition.yml
IreneMLC: If no dashboard is provided users who have this entity will see an error message on the entity overview @douglascamata

naxhh (approved):

entityBot (approved):

PR #136, opened 2021-03-29
Fix AWS tags and rule condition for K8s Deployments, Daemonsets and Containers

douglascamata:
### Relevant information

* We want to stop relying on `k8s.{deployment|daemonset}.createdAt` as it will very likely be transformed into a more generic tag named `k8s.createdAt` in the future.

* We're removing the `aws.ec2.instanceType` tag from all the k8s entities, as it didn't contain a value that actually reflected the EC2 instance type. 

* We're adding the `host.instanceType` tag to k8s entities that run on the same node as the integration that fetches data about them: pods and containers.

* We're also removing the `aws.region`  tag for  k8s objects that creates pods: as the `aws.region` comes from metadata decoration based on data reported by the integration pod, it will always contain the region of the integration pod itself, which is not always equal to the region where the **daemonsets** and **deployments** might be in. This happens because you can have a cluster with nodes on multiple regions.

### Checklist

 explanation above.


entityBot (approved):

javimb:
Looks good to me

IreneMLC (approved):

entityBot (approved):

definitions/infra-kubernetes_daemonset/definition.yml
```diff
@@ -28,4 +28,6 @@ synthesis:
     newrelic.agentVersion:
         multiValue: false
     aws.region:
+      multiValue: false
     aws.ec2.instanceType:
+      multiValue: false
```
naxhh, line 33: same question than in another pr. why do we think these ones will not be multi value?

douglascamata: We decided to actually drop this tag. But I'll explain it below.

I leave to think in reverse: what would be the implications of this tag being multi-value? This can only be answered by knowing the source of this tag.

This tag comes from data the agent fetches about the machines where it runs, when these machines are in AWS. They are not related to any monitored daemonset.

Then we get back to implications: you would see a daemonset instance with many different "instanceTypes", with values changing as the integration's pod get moved between nodes. This is extremely confusing and not helpful at all. So we don't want "many historical values", we want only the most recent.

entityBot (approved):

definitions/infra-kubernetes_daemonset/definition.yml
naxhh: Thanks! That makes a lot of sense

naxhh (approved):

entityBot (approved):

definitions/infra-kubernetes_pod/definition.yml
```diff
@@ -31,5 +31,6 @@ synthesis:
         multiValue: false
     newrelic.agentVersion:
         multiValue: false
+    host.instanceType:
     aws.region:
-    aws.ec2.instanceType:
+      multiValue: false
```
srvaroa, line 36: is the indentation right here? (I see it +2 chars deeper in L33)

douglascamata: Fixing this. The standard seems to be 2 spaces, right? Just confirming.

entityBot (approved):

entityBot (approved):

naxhh:
Merging since i understand the only thing to fix was the identation.

entityBot (approved):

PR #143, opened 2021-03-30
Add summary metrics, golden metrics and tags to k8s pods

douglascamata:
### Relevant information

Nothing special.

### Checklist

* [x I've read the guidelines and understand the acceptance criteria.
 explanation above.

### DIFF
--- a/definitions/infra-kubernetes_pod/golden_metrics.yml
+++ b/definitions/infra-kubernetes_pod/golden_metrics.yml
@@ -1,0 +1,21 @@
+statusOverTime:
+  title: Status changes over time
+  query:
+    select: count(*)
+    from: K8sPodSample
+    eventId: entityGuid
+    eventName: entityName
+podReadiness:
+  title: Readiness & status over time
+  query:
+    select: latest(isReady), latest(isScheduled)
+    from: K8sPodSample
+    eventId: entityGuid
+    eventName: entityName
+networkStats:
+  title: Network status
+  query:
+    select: max(net.rxBytesPerSecond), max(net.txBytesPerSecond), max(net.errorsPerSecond)
+    from: K8sPodSample
+    eventId: entityGuid
+    eventName: entityName
--- a/definitions/infra-kubernetes_pod/definition.yml
+++ b/definitions/infra-kubernetes_pod/definition.yml
@@ -3,7 +3,18 @@ type: KUBERNETES_POD
 configuration:
   entityExpirationTime: DAILY
   alertable: true
-
+goldenTags:
+  - k8s.status
+  - k8s.clusterName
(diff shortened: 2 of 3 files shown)


entityBot (approved):

definitions/infra-kubernetes_pod/golden_metrics.yml
```diff
@@ -0,0 +1,21 @@
...
+  title: Readiness & status over time
+  query:
+    select: latest(isReady), latest(isScheduled)
+    from: K8sPodSample
+    eventId: entityGuid
+    eventName: entityName
+networkStats:
+  title: Network status
+  query:
+    select: max(net.rxBytesPerSecond), max(net.txBytesPerSecond), max(net.errorsPerSecond)
```
naxhh, line 18: I've doubts with this specific one. Normally we have 1 expression here only. I'm unsure if golden signals will be able to handle this or if it evens makes sense to allow this. I'm opening a discussion with the team.

naxhh (commented):
aside of the golden metric with multiple expressions lgtm

definitions/infra-kubernetes_pod/golden_metrics.yml
srvaroa: +1

douglascamata: My intention was just to show these metrics in the MiniOverview in the same chart instead of having 3 charts. I can totally separate them if needed.

srvaroa (changes requested):
Setting request changes for now as we have to evaluate if the metric with 3 values is doable.

entityBot (approved):

definitions/infra-kubernetes_pod/golden_metrics.yml
rcrocker: TLDR - separate them

The way I like to think of the Golden Metrics is first as metrics - specifically, as something we can store atomically as a dimensional metric. Individually these are perfectly fine metrics. A 'group of three things' can't be stored as a dimensional metric.

I do see what you're trying to do though - you'd like to see all of these on the same chart. The team and I have been chatting about solving this problem by distinguishing between 'atomic' metrics and 'composition' metrics (that compose atomic metrics into a value) - that constraint of being a dimensional metric is limited to the atomic metrics. Having a composition of atomic metrics is a thing that I'd welcome as it solves many pragmatic problems - such as this case and error rate being examples.

entityBot (approved):

entityBot (approved):

entityBot (approved):

IreneMLC (approved):
Looks good!

entityBot (approved):

PR #148, opened 2021-03-31
Remove GM queries provider validation and add to query instead

IreneMLC:
### Relevant information

For GMs we were wrongly validating the name of the 'provider' in `queries` against our GM key regex, but the provider doesn't need to have the same restrictions (i.e. no dashes). 

This will fix a [PR that is failing](https://github.com/newrelic-experimental/entity-synthesis-definitions/pull/145) because of the wrong validation.


entityBot (approved):

naxhh (approved):
thanks!

PR #150, opened 2021-03-31
Add multiple synth rules for ext-service

IreneMLC:
### Relevant information

We want to synthesize ext-service entities from telemetry that contains any of the following attributes:

- service.name
- service_name
- serviceName


srvaroa (approved):
Thanks!

PR #159, opened 2021-04-01
Added golden metrics for multiple open instrumentation providers (ext-service)

jonatns:
### Relevant information

Added golden metrics for multiple open instrumentation providers under ext-service such as opentelemetry, kamon-agent and micrometer.

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-service/definition.yml
+++ b/definitions/ext-service/definition.yml
@@ -2,7 +2,28 @@ domain: EXT
 type: SERVICE
 synthesis:
   rules:
-    # opentelemetry provider
+    # telemetry with service.name attribute
+    - identifier: service.name
+      name: service.name
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: service.name
+
+    # telemetry with service_name attribute
+    - identifier: service_name
+      name: service_name
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: service_name
+
+    # telemetry with serviceName attribute
+    - identifier: serviceName
+      name: serviceName
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: serviceName
+
+    # telemetry from opentelemetry provider
     - identifier: service.name
       name: service.name
       encodeIdentifierInGUID: true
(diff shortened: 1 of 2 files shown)


entityBot (approved):

entityBot (approved):

definitions/ext-service/definition.yml
```diff
@@ -8,18 +8,59 @@ synthesis:
...
+
+    # telemetry from kamon-agent provider
+    - identifier: service.name
+      name: service.name
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: instrumentation.provider
+          value: kamon-agent
+      tags:
+        instrumentation.provider:
```
IreneMLC, line 48: This tag is already added by default I think ( @srvaroa)

definitions/ext-service/golden_metrics.yml
```diff
@@ -0,0 +1,38 @@
...
+      from: Metric
+    micrometer:
+      select: filter(sum(http.server.requests), where exception IS NOT NULL and exception != 'None') / sum(http.server.requests)
+      from: Metric
+responseTimeMs:
+  title: Response time (ms)
+  queries:
+    opentelemetry:
+      select: average(duration.ms)
+      from: Span
```
IreneMLC, line 31: Do we know if spans are ok for Golden signals @naxhh ?

naxhh: I can't recall, i'm guessing no but ping @rcrocker

entityBot (approved):

entityBot (approved):

definitions/ext-service/golden_metrics.yml
rcrocker: They should be ok.

IreneMLC: Thanks!

IreneMLC (approved):

entityBot (approved):

definitions/ext-service/definition.yml
srvaroa: it should be, yes

definitions/ext-service/definition.yml
```diff
@@ -8,18 +8,59 @@ synthesis:
...
+
     # telemetry with serviceName attribute
     - identifier: serviceName
       name: serviceName
       encodeIdentifierInGUID: true
       conditions:
         - attribute: serviceName
+
+    # telemetry from opentelemetry provider
+    - identifier: service.name
```
srvaroa, line 27: This is already being matched in the rule defined in L6.  I'm actually wondering if we should remove that one, and keep this one which is more specific.  From what I know the original intent of supporting `service.name` was OTEL so it allows us to match with more precission.  We were already seeing that a raw `service.name` match was too broad.

definitions/ext-service/definition.yml
naxhh, line 48:
```suggestion
```

PR #165, opened 2021-04-09
Improve tags documentation

IreneMLC:
### Relevant information

- Explain that we add any attributes prefixed with `tags.` as entity tags.
- Explain that we prefer these entity tags over rule-defined tags in case of collision.

We could also add here the list of reserved tags that only we can use if we want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,16 @@ We support the following conditions over telemetry attributes and their values:
 
 ##### Tags
 
+When synthesizing an entitym we add some tags to it by default based on telemetry attributes and others based on the entity synthesis rules.
+
+###### Default tags
+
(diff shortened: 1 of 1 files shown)


entityBot (approved):

cianBuckley (approved):

README.md
```diff
@@ -158,6 +158,16 @@ We support the following conditions over telemetry attributes and their values:
 
 ##### Tags
 
+When synthesizing an entitym we add some tags to it by default based on telemetry attributes and others based on the entity synthesis rules.
+
+###### Default tags
```
srvaroa, line 163: Mention that some are reserved? (e.g. name, accountId, accountName)?

IreneMLC: Sure, I wasn't sure if we wanted to add them already (I'll add them in another PR though, so I can merge this one to try and re-trigger the bot).

entityBot (approved):

PR #177, opened 2021-04-14
Remove node and namespace from pod summary metrics

douglascamata:
### Relevant information

We're trying to make it less crowded. Both these two tags can be used as filters in the UX.

### Checklist

 explanation above.


entityBot (approved):

IreneMLC (approved):
🧹

PR #185, opened 2021-04-20
[BOND-4026] add aws ecs container instance entity 

mkrichevsky:
### Relevant information

We're adding new entity type to cloud integrations https://newrelic.atlassian.net/browse/BOND-4026

### Checklist

 explanation above.


entityBot (approved):

definitions/infra-awsecsinstance/definition.yml
```diff
@@ -0,0 +1,19 @@
+domain: INFRA
+type: AWSECSCONTAINERINSTANCE
```
IreneMLC, line 2: The folder name needs to be {domain}-{type} so in your case `infra-awsecscontainerinstance` I think

entityBot (approved):

definitions/infra-awsecsinstance/definition.yml
mkrichevsky: Thank you! Missed that part. Pushed fix

IreneMLC (approved):
Looks good, thanks!

definitions/infra-awsecscontainerinstance/golden_metrics.yml
```diff
@@ -0,0 +1,16 @@
+registeredCpu:
+  title: Registered CPUs
+  query:
+    select: average(provider.registeredCpu)
+    from: ComputeSample
```
naxhh, line 5: just curious but is there any reason to use sample in the golden metrics and Metric in the summaries?

naxhh (approved):

definitions/infra-awsecscontainerinstance/golden_metrics.yml
mkrichevsky: I followed the same approach used in `infra-awsecsservice` and `infra-awsecscluster`. Seems like this approach is used in all our cloud integrations. If you think I should follow different approach, I can update with suggested changes

naxhh: no its ok! I just wanted to know if there was a specific reason for it :D

naxhh:
@mkrichevsky should be in staging now! It will go to prod when we issue a release.

I'm not sure if for those types work in the UI needs to be done in order to display them (but that's probably true)

PR #190, opened 2021-04-23
Ensure that telemetry.sdk.language is set

srvaroa:

### Relevant information

This tag seems to be missing from some entities.

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-service/definition.yml
+++ b/definitions/ext-service/definition.yml
@@ -8,6 +8,12 @@ synthesis:
       encodeIdentifierInGUID: true
       conditions:
         - attribute: service_name
+      tags:
+        telemetry.sdk.name:
+          entityTagName: instrumentation.provider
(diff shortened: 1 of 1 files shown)


entityBot (approved):

naxhh (approved):
LGTM but do they all really have this tag?

srvaroa:
Not necessarily, but if they do we should copy it.

definitions/ext-service/definition.yml
```diff
@@ -8,13 +8,25 @@ synthesis:
       encodeIdentifierInGUID: true
       conditions:
         - attribute: service_name
+      tags:
+        telemetry.sdk.name:
+          entityTagName: instrumentation.provider
```
naxhh, line 13: we can clean it up later but i don't think we should try to override provider in these. Since they already bring a provider.

PR #193, opened 2021-04-26
adding flow device definition

thezackm:
### Relevant information

adding definition for flow devices from Kentik

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-flow_device/summary_metrics.yml
+++ b/definitions/ext-flow_device/summary_metrics.yml
@@ -1,0 +1,33 @@
+endpoints:
+  title: Endpoints
+  unit: COUNT
+  query:
+    select: uniqueCount(ip_address)
+    from: Metric
+    where: "metricName LIKE 'kentik.rollup.bytes.%'"
+    eventId: entity.guid
+
+conversations:
+  title: Conversations
+  unit: COUNT
+  query:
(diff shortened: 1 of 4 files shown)


srvaroa (approved):
Thanks!

IreneMLC (approved):
Looks good, thanks!

definitions/ext-flow_device/summary_metrics.yml
```diff
@@ -0,0 +1,33 @@
...
+    select: uniqueCount(ip_address)
+    from: Metric
+    where: "metricName LIKE 'kentik.rollup.bytes.%'"
+    eventId: entity.guid
+
+conversations:
+  title: Conversations
+  unit: COUNT
+  query:
+    select: uniqueCount(tuple(ip_address,port))
```
naxhh, line 14: unsure if this will work out of the box in summaries. we should check

thezackm: it seems to be working right now in the [ext-vpc_network](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/b972d23d8ab26cec93dfc77a0ea5ae8aacbeb3fd/definitions/ext-vpc_network/summary_metrics.yml#L9-L16) entities that are live

PR #202, opened 2021-05-05
Add instrumentation.provider=opentelemetry to ext-service definitions.

a-feld:
### Relevant information

This PR stamps an entity guid when instrumentation.provider is opentelemetry

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-service/definition.yml
+++ b/definitions/ext-service/definition.yml
@@ -42,6 +42,18 @@ synthesis:
           entityTagName: language
         telemetry.sdk.version:
 
+    # telemetry from opentelemetry provider
+    - identifier: service.name
+      name: service.name
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: instrumentation.provider
+          value: opentelemetry
+      tags:
+        telemetry.sdk.language:
+          entityTagName: language
+        telemetry.sdk.version:
+
     # telemetry from kamon-agent provider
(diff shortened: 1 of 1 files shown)


entityBot (approved):

entityBot (approved):

srvaroa (approved):

definitions/ext-service/definition.yml
```diff
@@ -42,6 +42,18 @@ synthesis:
           entityTagName: language
         telemetry.sdk.version:
 
+    # telemetry from opentelemetry provider
+    - identifier: service.name
+      name: service.name
+      encodeIdentifierInGUID: true
+      conditions:
```
naxhh, line 49: will this one generate the same telemetry as the one above this rule?

srvaroa: No, the tag translation happens after there is a match

naxhh: I do agree on that. But do we need to have different Golden Metrics and summary metrics for this two? if that's the case having the same instrumentation.provider will be an issue.

If we don't need to, this is fine as it is

srvaroa: I don't think so because both are open telemetry so I'd expect the attributes to be the same (reinforcing this: we restored this rule because we noticed some entities missing, so the GMs were showing correctly already).

PR #204, opened 2021-05-10
Fix AWS Golden Metrics

jlegoff:
### Relevant information

Context: infrastructure can send metrics either using Samples or using Dimensional Metrics. The shim normally handles the conversion transparently, but the golden metrics harvester does not implement the shim. Because of that, we need to have golden metrics both in samples and DMs. In order to do that, we define a default query (with provider 'aws') and a sample version (with provider 'awsSample'). 

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-awsapigatewayapi/golden_metrics.yml
+++ b/definitions/infra-awsapigatewayapi/golden_metrics.yml
@@ -16,7 +16,7 @@ errors4XxAnd5Xx:
   title: Errors (4xx and 5xx)
   queries:
     aws:
-      select: sum(aws.apigateway.5XXError.byApi) + sum(aws.apigateway.4XXError.byApi)
+      select: sum(`aws.apigateway.5XXError.byApi`) + sum(`aws.apigateway.4XXError.byApi`)
       from: Metric
       eventId: entity.guid
       eventName: entity.name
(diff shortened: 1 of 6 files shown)


definitions/infra-awsapigatewayapi/golden_metrics.yml
```diff
@@ -1,24 +1,42 @@
...
   title: Errors (4xx and 5xx)
-  query:
-    select: sum((`provider.5xxError.Sum` OR 0) + (`provider.4xxError.Sum` OR 0))
-    from: ApiGatewaySample
-    where: provider='ApiGatewayApi'
-    eventId: entityGuid
-    eventName: entityName
+  queries:
+    aws:
+      select: sum(aws.apigateway.5XXError.byApi) + sum(aws.apigateway.4XXError.byApi)
```
naxhh, line 19: bot complains because these should be escaped

```suggestion
      select: sum(`aws.apigateway.5XXError.byApi`) + sum(`aws.apigateway.4XXError.byApi`)
```

entityBot (approved):

IreneMLC (approved):

srvaroa (approved):

PR #205, opened 2021-05-11
Add rule for EXT-SERVICE from JFR-Uploader

srvaroa:

### Relevant information

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-service/definition.yml
+++ b/definitions/ext-service/definition.yml
@@ -28,6 +28,20 @@ synthesis:
           entityTagName: language
         telemetry.sdk.version:
 
+    # telemetry from JFR uploader
+    - identifier: service.name
+      name: service.name
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: telemetry.sdk.name
+          value: JFR-Uploader
+      tags:
+        telemetry.sdk.name:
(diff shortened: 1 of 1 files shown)


entityBot (approved):

definitions/ext-service/definition.yml
```diff
@@ -28,6 +28,20 @@ synthesis:
           entityTagName: language
         telemetry.sdk.version:
 
+    # telemetry from JFR uploader
+    - identifier: service.name
```
naxhh, line 32: nit but should we put this near the other JFR-UPLOADER rule?

naxhh (approved):

definitions/ext-service/definition.yml
srvaroa: oh there is one already!

naxhh: yeah but is by `instrumentation.provider` would that match the same entities as this one?

srvaroa: Yeah, it was enough with instrumenation.provider.  The issue was that we had disabled EXT-SERVICE in the ingestion pipeline that processed the JFR-Uploader events.

PR #206, opened 2021-05-11
Jlegoff/fix azure golden metrics

jlegoff:
### Relevant information

As a follow up to https://github.com/newrelic-experimental/entity-synthesis-definitions/pull/204, here are the updated definitions for Azure entities

### Checklist

 explanation above.


naxhh:
blocking this for now, since we reverted the AWS ones.

PR #217, opened 2021-05-20
Fix EXT-ROUTER definition

IreneMLC:
### Relevant information

- Add back the correct attribute for kentik rules (`instrumentation.provider` -> `provider`)
- Fix tags for `mikrotik` rules


entityBot (approved):

IreneMLC:
I'm going ahead an merging this since it can break kentik entities, if you see any issues let me know and I'll open a follow-up

PR #220, opened 2021-05-26
Ruairi/gatsby build entity

rudouglas:
### Relevant information

We use Gatsby for the Docs site etc. and we instrument the build steps using Jaeger and OpenTelemetry

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-gatsby_build/definition.yml
+++ b/definitions/ext-gatsby_build/definition.yml
@@ -6,8 +6,7 @@ synthesis:
   identifier: gatsbySite
   encodeIdentifierInGUID: true
   conditions:
-    - attribute: service.name
-      value: gatsby
+    - attribute: gatsbySite
 
   tags:
     site:
--- a/definitions/ext-gatsby_build/definition.yml
+++ b/definitions/ext-gatsby_build/definition.yml
@@ -1,19 +1,16 @@
 domain: EXT
-# Derived from the Fastly log ingestion, a datacenter is a Point of Presence
-# https://docs.fastly.com/en/guides/fastly-pop-locations
 type: GATSBY_BUILD
 
 synthesis:
   name: gatsbySite
   identifier: gatsbySite
-  encodeIdentifierInGUID: false
+  encodeIdentifierInGUID: true
   conditions:
-    - attribute: serviceName
+    - attribute: service.name
       value: gatsby
 
   tags:
     site:
-    build:
 
 configuration:
   entityExpirationTime: EIGHT_DAYS
--- a/definitions/ext-gatsby_build/definition.yml
+++ b/definitions/ext-gatsby_build/definition.yml
@@ -1,0 +1,20 @@
+domain: EXT
+# Derived from the Fastly log ingestion, a datacenter is a Point of Presence
+# https://docs.fastly.com/en/guides/fastly-pop-locations
+type: GATSBY_BUILD
+
+synthesis:
+  name: gatsby_build
+  identifier: gatsby_build
+  encodeIdentifierInGUID: false
+  conditions:
+    - attribute: serviceName
+      value: gatsby
+
+  tags:
+    site:
+    build:
+
+configuration:
+  entityExpirationTime: EIGHT_DAYS
+  alertable: true
--- a/definitions/ext-gatsby_build/definition.yml
+++ b/definitions/ext-gatsby_build/definition.yml
@@ -4,8 +4,8 @@ domain: EXT
 type: GATSBY_BUILD
 
 synthesis:
-  name: gatsby_build
-  identifier: gatsby_build
+  name: gatsbySite
+  identifier: gatsbySite
   encodeIdentifierInGUID: false
   conditions:
     - attribute: serviceName


entityBot (approved):

definitions/ext-gatsby_build/definition.yml
```diff
@@ -0,0 +1,20 @@
+domain: EXT
+# Derived from the Fastly log ingestion, a datacenter is a Point of Presence
```
srvaroa, line 2: The comment is not relevant for this type, right?

definitions/ext-gatsby_build/definition.yml
```diff
@@ -0,0 +1,20 @@
...
+# https://docs.fastly.com/en/guides/fastly-pop-locations
+type: GATSBY_BUILD
+
+synthesis:
+  name: gatsbySite
+  identifier: gatsbySite
+  encodeIdentifierInGUID: false
```
srvaroa, line 9: We probably want this to true, I assume the gatsbySite is likely to be potentially over 100 chars?

definitions/ext-gatsby_build/definition.yml
```diff
@@ -0,0 +1,20 @@
...
+  conditions:
+    - attribute: serviceName
```
srvaroa, line 11: Do you have a sample of telemetry that would match for these? (I'm interested in seeing if we can find a more specific attribute to identify builds).

definitions/ext-gatsby_build/definition.yml
rudouglas: Oh yes good call, I had copied this from a template :D

definitions/ext-gatsby_build/definition.yml
rudouglas: I'm not sure if it ever would be, this is set manually, unless i'm misunderstanding the use of `identifier`

definitions/ext-gatsby_build/definition.yml
rudouglas: Ye sure here is one of the spans @srvaroa  https://staging.one.nr/0GbRmvbenjy

srvaroa: I get this service.name in that span:

image

So that would not be matching this rule, right?

definitions/ext-gatsby_build/definition.yml
rudouglas: Ok cool i'll set `encodeIdentifierInGUID: true`. Ye so after sending those traces, I wanted to refactor the attributes a little, so thought `gatsbySite` might be a good unique identifier, i'll run a trace now to test that first though

definitions/ext-gatsby_build/definition.yml
rudouglas: aha ok ye apologies i'll run another test and refactor those suggestions 😬

rudouglas: Ok this is the updated span, so `service.name` and `gatsbySite` should be the best way i think?

entityBot (approved):

definitions/ext-gatsby_build/definition.yml
rudouglas: I've updated but open to your advice @srvaroa :D

cianBuckley (approved):

entityBot (approved):

definitions/ext-gatsby_build/definition.yml
IreneMLC: This telemetry is already associated with an ext-service entity though (from the broad rule I think), so we'd have conflicts there right @srvaroa?

IreneMLC (changes requested):
The telemetry used for this entity type also matches ext-service, so let's check that before merging this

definitions/ext-gatsby_build/definition.yml
srvaroa: Hi @rudouglas , following from @IreneMLC 's comment. We need to support [a rule](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/main/definitions/ext-service/definition.yml#L118) to match EXT-SERVICE entities from OpenTelemetry data, which is arguably too broad.  The rule that you propose here (which would otherwise be sensible) will be problematic in the current implementation as we'll get two potential matches (EXT-SERVICE vs. EXT-GATSBY_BUILD).

We can make a change in the rule engine to work around this and put less priority on EXT-SERVICE, but this involves adding exceptions to the matching process (with code changes & deploys) so I wanted to check if we have other alternatives before going down that route.

Is there a chance that your telemetry does not send the `service.name` attribute at all (so we'd write the rule matching on a different attribute, and avoid the overlap with EXT-SERVICE)? (I understand this may not be possible if you don't control the integration, but just wanted to check first).

rudouglas: Thanks @IreneMLC for your help, that does make sense, the service itself does need a `service.name` in order to function, but I can tag it with anything else we like, so i'm wondering if I used `gatsbyService: true` like in the following example https://staging.one.nr/0znQxNW9GRV or some variation on that. As I can decide what tags we add I assume we can work around this without you having to refactor the `EXT-SERVICE`, unless i'm wildly mistaken :P @srvaroa

srvaroa: Having the `service.name` is tricky because it will trigger the EXT-SERVICE rule just by being there, but I actually think we have a 3rd option if you have control over the telemetry, which is setting an `entity.type` attribute to `GATSBY_BUILD`.  This will make the rule engine favour that type before it considers EXT-SERVICE.

The changes would be basically

* Add that `entity.type` with value `GATSBY_BUILD` to your telemetry.
* Update the synthesis rule to match on something else than `service.name`, like:

```
 conditions:
          - attribute: gatsbySite
```

Would that work? (I would like to verify this once we merge the rule, but I think this should give us the desired outcome)

rudouglas: ok cool, i tried to set `entity.type` as a tag which didn't seem to overwrite it in NRDB, for context here we are using the new `otlp-native-ingest` container so i'm wondering if that's what's setting the `entity.type`:
https://docs-preview.newrelic.com/docs/otlp-native-ingest/

srvaroa: I think that's normal, the GATSBY_BUILD type doesn't exist right now so the match will land in EXT-SERVICE (due to the  `service.name` attributes). 

Let me merge the PR so we can see if this works once the type is present.

definitions/ext-gatsby_build/definition.yml
```diff
@@ -0,0 +1,17 @@
+domain: EXT
+type: GATSBY_BUILD
+
+synthesis:
+  name: gatsbySite
+  identifier: gatsbySite
+  encodeIdentifierInGUID: true
+  conditions:
+    - attribute: service.name
```
srvaroa, line 9: Can you update this to:

```
conditions:
  - attribute: gatsbySite
```

And no value? (this should fully avoid the EXT-SERVICE rule)

definitions/ext-gatsby_build/definition.yml
srvaroa: I noticed the condition is missing update, but I can merge after that.

entityBot (approved):

definitions/ext-gatsby_build/definition.yml
rudouglas: oooh nice @srvaroa i made that change there :D 🤞

definitions/ext-gatsby_build/definition.yml
rudouglas: done 👯

entityBot (approved):

srvaroa (approved):

srvaroa:
Waiting for @IreneMLC's OK but I think we should be able to try this out today.

IreneMLC (approved):
Looks good, thanks!

PR #232, opened 2021-06-04
Define definition of synthetic relationships 

cianbuckley:
~I want to use this as a point of discussion. Not looking to merge~ 

I'd like to commit the schema and validation files for relationships. All relationships defined (APM, BROWSER, INFRA) are merely sample relationships and used to demonstrate usage. ~I'll delete those before merging but leave there for reference and further discussion up to point of merge~ I'm thinking i'd like to merge through these relationships as a starting point to help us build the pipelines. We can remove them later. opinions?

### Relevant information

### Checklist

 explanation above.


entityBot (approved):

definitions/browser-application/definition.yml
```diff
@@ -7,3 +7,21 @@ compositeMetrics:
 configuration:
   entityExpirationTime: QUARTERLY
   alertable: true
+
+# Select entityGuid, requestUrl from AjaxRequest where requestUrl is not null
+relationships:
```
IreneMLC, line 12: This relationship section could grow a bit right? (CALLS, IS, etc) would it make sense to keep it as a separate file on the folder? with maybe just some configuration on the main definition file.

cianBuckley: ya thats an option!

definitions/infra-host/definition.yml
```diff
@@ -21,3 +21,17 @@ compositeMetrics:
 configuration:
   entityExpirationTime: DAILY
   alertable: true
+
+  #  SELECT entityGuid (target), nodeName (source) FRom Span where nodeName is not null
+  #  existing relationship
```
cianBuckley, line 26: The query is this `SELECT entityGuid as target, nodeName as source FRom Span where nodeName is not null` the target in this case is where the telemetry is coming from, but to me it makes most sense to declare the rule on the target because it can have a relationship between multiple types without having to be declared multiple times

As a consequence, navigation to find all rules with source|target as X will be more difficult.

definitions/apm-application/definition.yml
```diff
@@ -11,3 +11,22 @@ compositeMetrics:
 configuration:
   entityExpirationTime: QUARTERLY
   alertable: true
+
+relationships:
+  # Select entityGuid, peer.hostname from Span where peer.hostname is not null and category =
+  # 'datastore'
+  # ExtService in this case is a DB
+  - provider: newrelic
+    eventType: Span
```
srvaroa, line 20: Would it make sense to make `eventType` an array?

definitions/apm-application/definition.yml
```diff
@@ -11,3 +11,22 @@ compositeMetrics:
...
+    source:
+      guid: entityGuid
```
srvaroa, line 22: I'd assume that relationship synthesis would happen after entity synthesis, when the `entity.guid` is already set (either by synthesis, or by another ingestion service).  Taking that into account, could we just remove `source` altogether? (e.g. assuming the guid is at `entity.guid`).

srvaroa: That said, we most likely want to filter on  specific events (e.g. "apply this rule when the event is of type EXT-SERVICE). But I think the guid field should be taken by convention.

cianBuckley: my original draft was as you said, but then while defining infra host HOSTS X entity, I realised I wanted to change direction of the relationship i was creating. 

Saying that, we could still take the guid field by convention and force to specify only direction

definitions/apm-application/definition.yml
cianBuckley: maybe? I'm not sure. The attributes would possibly/likely change as well so i'm unsure of how useful it would really be?

definitions/infra-host/definition.yml
cianBuckley: no harm either right? Also better mimics what nrql can do

definitions/apm-application/definition.yml
cianBuckley: no harm either right? Also better mimics what nrql can do

entityBot (approved):

entityBot (approved):

entityBot (approved):

definitions/browser-application/definition.yml
cianBuckley: done! Although I'll skip any config in definition file

definitions/apm-application/definition.yml
cianBuckley: done!

definitions/apm-application/definition.yml
cianBuckley: done!

entityBot (approved):

definitions/apm-application/relationships.yml
```diff
@@ -0,0 +1,30 @@
...
+  - provider: newrelic
+    eventType: Span
+    rootEntityNodeType: SOURCE
+    relationship: CALLS
+    relatedEntity:
+      domain: EXT
+      type: SERVICE
+      domainId: peer.hostname
+      encodeIdentifierInGUID: true
+    conditions:
```
cianBuckley, line 14: multi conditions OR | AND

definitions/apm-application/relationships.yml
```diff
@@ -0,0 +1,30 @@
+relationships:
+  # Select entityGuid, peer.hostname from Span where peer.hostname is not null and category =
+  # 'datastore'
+  # ExtService in this case is a DB
+  - provider: newrelic
+    eventType: Span
+    rootEntityNodeType: SOURCE
+    relationship: CALLS
+    relatedEntity:
```
cianBuckley, line 9: mark as source|target - remove rootEntityNodeType

definitions/apm-application/relationships.yml
```diff
@@ -0,0 +1,30 @@
+relationships:
+  # Select entityGuid, peer.hostname from Span where peer.hostname is not null and category =
+  # 'datastore'
+  # ExtService in this case is a DB
+  - provider: newrelic
```
cianBuckley, line 5: change provider - instrumentation.provider

definitions/apm-application/relationships.yml
```diff
@@ -0,0 +1,30 @@
+relationships:
+  # Select entityGuid, peer.hostname from Span where peer.hostname is not null and category =
```
cianBuckley, line 2: Rule for location of definition is: Define it in the file where the data point is originating

cianBuckley: Add documentation

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

definitions/browser-application/definition.yml
IreneMLC: thanks!

definitions/apm-application/relationships.yml
```diff
@@ -0,0 +1,26 @@
...
+  # 'datastore'
+  # ExtService in this case is a DB
+  - instrumentation.provider: newrelic
```
naxhh, line 5: super mega nit: newRelic is the default one

docs/example-relationship-definition.yml
```diff
@@ -0,0 +1,54 @@
+# a list of relationships defined in a file under the entity in which the telemetry is sourced
```
IreneMLC, line 1: In the real `relationships.yml` files you are adding a query on top to see the data, that seems useful, should we suggest that people add it?

definitions/browser-application/relationships.yml
```diff
@@ -0,0 +1,10 @@
+# Select entityGuid, requestUrl from AjaxRequest where requestUrl is not null
+relationships:
+  - instrumentation.provider: newrelic
+    eventType: AjaxRequest
+    relationship: CALLS
+    target:
+      domain: EXT
+      type: SERVICE
+      domainId: requestUrl
+      processor: hostExtractor
```
naxhh, line 10: I'm wondering if the domainId should be an object. Since the processor is needed in order to obtain it.
Like:
```
domainId:
  attribute: requestUrl
  processor: hostExtractor
```

We can keep the simpler form ` domainId: requestUrl` when no processor is rquired.
Thoughts?

definitions/infra-host/relationships.yml
```diff
@@ -0,0 +1,11 @@
+  #  SELECT entityGuid (target), nodeName (source) FRom Span where nodeName is not null
+  #  existing relationship
+relationships:
```
naxhh, line 3: should this one be on in ext-service. If i'm not wrong the nodeName is reported by the service right?

validator/schemas/relationship-schema-v1.json
```diff
@@ -0,0 +1,237 @@
...
+            "title": "Relationship Type",
+            "description" : "The relationship edge type you are looking to build between two entity nodes",
+            "examples" : ["CALLS", "HOSTS", "IS", "CONTAINS"],
+            "enum" : ["CALLS", "HOSTS", "IS", "CONTAINS"]
+          },
+          "conditions": {
+            "$id": "#/relationships/items/conditions",
+            "type": "array",
+            "minItems": 1,
+            "maxItems": 1,
```
IreneMLC, line 78: This will make us always request exactly 1 condition, but we said that they'd be optional for relationships right?

validator/schemas/relationship-schema-v1.json
```diff
@@ -0,0 +1,237 @@
...
+        "additionalProperties": false,
+        "properties": {
+          "instrumentation.provider": {
+            "$id": "#relationships/items/instrumentation.provider",
+            "type": "string",
+            "title": "Instrumentation Provider",
+            "description": "instrumention.provider agent or source from where the telemetry is coming",
+            "examples": ["newrelic"]
+          },
+          "eventType" : {
```
naxhh, line 59: wondering if this should be optional or not. Like synthesis definition?

validator/tools/validate_relationship_rules.js
```diff
@@ -0,0 +1,133 @@
+const utils = require('./utils')
```
IreneMLC, line 1: This is the same as `validate_rules.js` and not needed right?

validator/schemas/relationship-schema-v1.json
cianBuckley: It's still optional. This means if you add a condition, you can add max 1 min 1

definitions/apm-application/relationships.yml
cianBuckley: updated!

definitions/browser-application/relationships.yml
cianBuckley: ya that makes sense. it more clearly links the processor strictly to the domainID as actually intended

definitions/infra-host/relationships.yml
cianBuckley: well.. problem is if we do that, we would need to declare this rule in N places.

validator/schemas/relationship-schema-v1.json
cianBuckley: What would the default be?

validator/tools/validate_relationship_rules.js
cianBuckley: oops! thanks :D

docs/example-relationship-definition.yml
cianBuckley: ya! nice suggestion

entityBot (approved):

definitions/apm-application/relationships.yml
```diff
@@ -2,7 +2,7 @@ relationships:
   # Select entityGuid, peer.hostname from Span where peer.hostname is not null and category =
```
srvaroa, line 2: I would personally move the relationship synthesis definitions into `definitions.yml`.

definitions/apm-application/relationships.yml
srvaroa: Do we need them now?

definitions/apm-application/relationships.yml
```diff
@@ -0,0 +1,26 @@
...
+        value: datastore
+
+  # SELECT count(*) From Span facet request.domain where request.domain is not null and
+  # entityGuid is not null
+  - instrumentation.provider: newRelic
+    eventType: Span
+    relationship: CALLS
+    target:
+      domain: EXT
+      type: SERVICE
```
srvaroa, line 24: Is this rule actually wanted? AFAIK the entities that have the request.domain are those "external" pseudo entities that won't work until they are indexed.  If this is the case I'd remove this rule.

README.md
```diff
@@ -49,6 +49,7 @@ They must be placed inside the same folder as the entity definition and must be
 
 - Golden metrics: `golden_metrics.yml`
 - Summary metrics: `summary_metrics.yml `
+- Relationship Definitions: `relationships.yml `
```
srvaroa, line 52: Sorry i misplaced this comment elsewhere.  I think the relationship synthesis defs. are better suited for `definitions.yml`

definitions/infra-host/relationships.yml
srvaroa: yeah but I think you can't avoid it

Declaring it here implies that on the host telemetry you expect datapoints to contain attributes for *every* service that runs in that node. Hence, you have to alter the agent to pull those services and report them in separate attributes.  Which I think has obvious complication (e.g. if you have 2 services running in a host, the agent needs 2 attributes, and we need 2 rules to understand each).

I agree with @naxhh that this rule makes more sense on the service, as given a datapoint, there will always be a 1-1 relationship to a node (a single attribute).  If we have another type of entity that also can bring the nodeName, then yes, we need a rule there.

docs/example-relationship-definition.yml
srvaroa: +1 although I would probably use natural language to explan the rule (as the user may not necessarily understand the query approach)

docs/example-relationship-definition.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      # we have a predefined list of functions that you can call during synthesis which will help
+      # you extract or transform the domainID attribute value into something you're expecting. 
+      # as an example, take `hostExtractor` if your assigned domainID attribute value is
+      # https://myhost.com/bla/bla then the hostExtractor will return myhost.com ignoring the rest
+      processor: hostExtractor
+
+    # are there any filters you wish to apply. if so, you can match attributes values to what you
+    # expect here. eg attrA = 'foo' AND attrB = 'bar'
+    conditions:
```
srvaroa, line 39: If we plan to support multiple attributes, should we just remove the instrumentation.provider from above and put it here?

I can see some cases where `instrumentation.provider` alone will not be enough (e.g. Kentik), so we'll still need multiple conditions.

docs/example-relationship-definition.yml
```diff
@@ -0,0 +1,60 @@
...
+  # providing browsing information 
+  - instrumentation.provider: newrelic
+    # what NRDB event type should we filter telemetry from. eg Transaction, AjaxRequest, Span or a
+    # custom event type
+    eventType: Span
+    # there is a predefined set of relationship types. this is the connecting edge between the
+    # source and target nodes. eg CALLS, HOSTS, CONTAINS, IS
+    relationship: CALLS
+    # information of how to formulate the target node of the relationship 
+    target:
```
srvaroa, line 19: Source and target are mutually exclusive, so I'd show this as 

```
[source | target]:
  ...
```

Adapting the explanations in the comments.

validator/schemas/relationship-schema-v1.json
```diff
@@ -0,0 +1,237 @@
...
+                      "examples": [
+                        "value"
+                      ]
+                    }
+                  }
+                }
+              ]
+            }
+          },
+          "target": {
```
srvaroa, line 141: We should have target or source, not both, right?

definitions/apm-application/relationships.yml
cianBuckley: I thought we agreed to add and|or later?

definitions/infra-host/relationships.yml
cianBuckley: Considering this, we could likely remove the `source` node. No need for it if it's to always be the `entityGuid` of the defining entity type

definitions/apm-application/relationships.yml
cianBuckley: I've no strong opinion. It was originally there and suggested to move it out of there so i complied. Also, it follows convention as is.

definitions/apm-application/relationships.yml
cianBuckley: All 4 of these rules need to be properly revised and likely just removed.
Saying that, I've manually searched through and found some existing open telemetry entities using that as the domainID. I think it's good enough data to test the pipeline with as we build it out.

definitions/apm-application/relationships.yml
srvaroa: PRs > VC for this reason!

My preferences would be only one attribute. And, if we do have a real use case for N attributes, then only suppor AND acrosss all of them. Or is implemented by creating a copy of the rule  (this is what we do in entity synthesis, to avoid complicating the syntax).

I can see a case for going to N attributes with AND because that way we could just not have `instrumentation.provider`. 

But I'm reluctant to going beyond that because logic operators to the yaml it feels antipattern-y to me (e.g. yaml stops being configuration and starts begin a programming language).

docs/example-relationship-definition.yml
cianBuckley: I'll modify to support multiple attributes. 

IMO i prefer it root level. I think it's easier to explain it as it is other than requiring it as a condition

definitions/apm-application/relationships.yml
srvaroa: Yup, it would be interesting to know which type is being used in those cases (e.g. is it SERVICE or the THIRD_PARTY_SERVICE_ENTITY ?)

docs/example-relationship-definition.yml
srvaroa: It should be trivial-ish to move it later if we decide to so no problem with keeping it

entityBot (approved):

entityBot (approved):

definitions/apm-application/relationships.yml
cianBuckley: we don't have any actual entities with `THIRD_PARTY_SERVICE_ENTITY` so it would have to be `SERVICE`

docs/example-relationship-definition.yml
cianBuckley: done!

docs/example-relationship-definition.yml
cianBuckley: done!

validator/schemas/relationship-schema-v1.json
cianBuckley: removed min and max

definitions/apm-application/relationships.yml
cianBuckley: I agree with starting with support for just AND. However, multiple AND clauses would imply more filtering thus less processing so I'm all for permitting multiple off the bat. what you think

definitions/apm-application/relationships.yml
cianBuckley: Digging deeper though, none of these entities actually have telemetry right now. Something is not right.

Looking further still, all of the ext entities here are over a month old. Perhaps they are left over from when ext-service rules were too open

cianBuckley: Even though this rule will likely never actually match an existing target entity, it could still be handy for test. Ultimately, we don't want to generate or at least not show a relationship to a non existing entity. We can use this rule to test that case

cianBuckley: I'll leave a comment explaining

entityBot (approved):

definitions/browser-application/relationships.yml
cianBuckley: couldn't quite figure out how to make it play nice with both structures, but i modified domainID to be an object that accepts processor and attribute!

docs/example-relationship-definition.yml
IreneMLC: thanks!

definitions/apm-application/relationships.yml
IreneMLC: We didn't want to allow the `AND` on entity definitions, but here it should be fine right?

srvaroa: Yes, N match conditions that are all required should not be problematic.

definitions/apm-application/relationships.yml
```diff
@@ -0,0 +1,37 @@
+relationships:
+  # Select entityGuid, peer.hostname from Span where peer.hostname is not null and category =
+  # 'datastore'
+  # ExtService in this case is a DB
+  #
+  # TODO revise rule 
+  - instrumentation.provider: newRelic
```
srvaroa, line 7: For me this is an example of why it'd make sense to have these in the definitions file itself.  In cases like OTEL, the instrumentation.provider tag is not present in the original telemetry (we add a tag rule to copy it from telemetry.sdk.name) and there is a rule so I can imagine a user being confused by what to put here or where that value comes from.  Maybe worth checking with product to see which side of the tradeoff is better here.

PR #235, opened 2021-06-09
feat: adding ext-ups entity definition

thezackm:
### Relevant information

Adding definition for external UPS devices received from Kentik SNMP integration via Metrics API

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-ups/definition.yml
+++ b/definitions/ext-ups/definition.yml
@@ -1,0 +1,27 @@
+domain: EXT
+type: UPS
+synthesis:
+  name: device_name
+  identifier: device_name
+  encodeIdentifierInGUID: true
+
+  conditions:
+  - attribute: provider
+    value: kentik-ups
+
+  tags:
+    src_addr:
+      entityTagName: device_ip
+
+compositeMetrics:
+  goldenMetrics:
+  - golden_metrics.yml
+  summaryMetrics:
+  - summary_metrics.yml
+
+goldenTags:
+- device_ip
+
+dashboardTemplates:
+  newRelic:
+    template: dashboard.json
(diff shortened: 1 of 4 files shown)


entityBot (approved):

entityBot (approved):

definitions/ext-ups/definition.yml
```diff
@@ -0,0 +1,27 @@
...
+  goldenMetrics:
+  - golden_metrics.yml
+  summaryMetrics:
+  - summary_metrics.yml
+
+goldenTags:
+- device_ip
+
+dashboardTemplates:
+  newRelic:
```
IreneMLC, line 26: Same as with the other PR, I think that you could use `kentik` here instead (that's the value of the `instrumentation.provider` for this telemetry right?) that way it'd be easier for people to extend this definition for other providers, what do you think?

IreneMLC (approved):

definitions/ext-ups/definition.yml
thezackm: [same as the other PR] that's a good callout - I was thinking to wait until the new standard that @srvaroa and folks are working on is in place for using `instrumentation.provider` ++ ??? and then I could push an update to all of the Kentik entities at once. 

Would you rather I submit the change to `kentik` on this PR and then update later?

naxhh (approved):
great to see this new types!

entityBot (approved):

definitions/ext-ups/definition.yml
thezackm: per conversation on PR #236 - updating to `kentik` to match now and we will address later once new standard is in place

entityBot (approved):

PR #236, opened 2021-06-09
feat: adding ext-firewall entity definition

thezackm:
### Relevant information

Adding `ext-firewall` entity definition from Kentik SNMP integration using Metrics API. 

**NOTE: This definition would only apply to a single vendor's data (Fortigate); future vendors will need to be added as additional sources much like the `ext-router` definition has today**

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-firewall/definition.yml
+++ b/definitions/ext-firewall/definition.yml
@@ -1,0 +1,30 @@
+domain: EXT
+# Firewall devices from Kentik using Metric API
+# Note this initial definition is limited to only Fortigate Firewalls using the FORTINET-FORTIGATE-MIB
+# Future firewall entities will need to be isolated in the definition
+type: FIREWALL
+synthesis:
+  name: device_name
+  identifier: device_name
+  encodeIdentifierInGUID: true
+
+  conditions:
+  - attribute: provider
+    value: kentik-firewall
+
+  tags:
+    src_addr:
+      entityTagName: device_ip
+
+compositeMetrics:
+  goldenMetrics:
+  - golden_metrics.yml
+  summaryMetrics:
+  - summary_metrics.yml
+
+goldenTags:
+- device_ip
+
+dashboardTemplates:
+  newRelic:
+    template: dashboard.json
--- a/definitions/ext-firewall/definition.yml
+++ b/definitions/ext-firewall/definition.yml
@@ -26,5 +26,5 @@ goldenTags:
 - device_ip
 
 dashboardTemplates:
-  newRelic:
+  kentik:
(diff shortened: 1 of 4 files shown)


entityBot (approved):

IreneMLC (approved):

definitions/ext-firewall/definition.yml
```diff
@@ -0,0 +1,30 @@
...
+  goldenMetrics:
+  - golden_metrics.yml
+  summaryMetrics:
+  - summary_metrics.yml
+
+goldenTags:
+- device_ip
+
+dashboardTemplates:
+  newRelic:
```
IreneMLC, line 29: You could use `kentik` here instead (that's the value of the `instrumentation.provider` for this telemetry right?) that way it'd be easier for people to extend this definition for other providers

thezackm: that's a good callout - I was thinking to wait until the new standard that @srvaroa and folks are working on is in place for using `instrumentation.provider` ++ ??? and then I could push an update to all of the Kentik entities at once. 

Would you rather I submit the change to `kentik` on this PR and then update later?

definitions/ext-firewall/definition.yml
```diff
@@ -0,0 +1,30 @@
...
+  identifier: device_name
+  encodeIdentifierInGUID: true
+
+  conditions:
+  - attribute: provider
+    value: kentik-firewall
+
+  tags:
+    src_addr:
+      entityTagName: device_ip
```
naxhh, line 17: what is this value? who is calling the firewall?

what is the expected amount of values here? Because I would be a bit wary of having all the internet ip's in this tag

definitions/ext-firewall/definition.yml
naxhh: Yeah I think adding here kentik would be less confusing in the future.

When we adde those changes we still will need to do some batch updates here anyway.

naxhh (approved):

definitions/ext-firewall/definition.yml
thezackm: this is the IP of the firewall itself; should be fairly static and in almost all cases would never change

entityBot (approved):

definitions/ext-firewall/definition.yml
thezackm: change added; thanks!

IreneMLC: thanks!

entityBot (approved):

PR #241, opened 2021-06-14
Add support for javacsript video players

cjmccarthy:
### Relevant information

We currently have a set of video player trackers ([ex: html5](https://github.com/newrelic/video-html5-js)) that collect event-based telemetry for video players. We'd like to synthesize entities based off of this data to make it much easier for customers to visualize and reason about their player performance, as well as start removing duplicate work and being taking advantage of platform features available for entities.

There are some oddities here that I'd like some advice on:

- The events are all sent via PageAction from our Browser API. Because of this, they all contain the Browser entityGuid, but we would like to synthesize the Video Player as a unique and separate entity (which would ideally eventually have a synthesized relationship to its originating Browser). I've use the browser appId as the `identifier` for now - it seems like a hack but we don't currently have another way of uniquely identifying the Video Player other than that and the Browser entityGuid which I'm assuming we cannot (or, at least, should not) re-use. I'd love some way to do a composite `identifier` - like re-hashing the browser GUID with the Player Type (html5, jwplayer, etc).
- Above point basically repeated for name. I'm re-using the Browser App Name, but that doesn't seem like as big of an issue. It would be neat to be able to define a rule to prepend/append a substring to an attribute for the name though (ex: `appName + ' - ' + playerName`)
- It's not clear what NRQL syntax is supported in the `select` clause for the derived metrics. We require the use of `filter(...)` to be able to calculate useful metrics, so I wanted to check if that's supported. Apologies if I missed that in the docs, but I couldn't find info on that.

This is my first swing at creating new entity types so any general help and guidance are greatly appreciated 😄  

### Checklist

 explanation above.


entityBot (approved):

entityBot (approved):

naxhh:
> The events are all sent via PageAction from our Browser API. Because of this, they all contain the Browser entityGuid, but we would like to synthesize the Video Player

One core rule is that one datapoint must be related to one entity only. This is because the metric gets the guid stamped into it. And all the entities ecosystem relies on that guid being on telemetry (golden metrics, summary metrics, golden signals, dashboards, etc)

srvaroa:
As @naxhh the main issue is that telemetry is also linked to an entity (the browser one) so the synthesis engine will skip this telemetry.

That said, I think this is probably a good case for relationship synthesis though, we're actively working on this at the moment.    The idea being that we define a relationship synthesis rule very similar to the one you provide here.  The engine would see that the telemetry is linked to a browser entity, and decide that there should be another entity (the video player) and a relationship between both.

If that makes sense for you I'd like to take this case as a sample for the new development we're working on.

cjmccarthy:
I was worried we might run into something like that!

@srvaroa I think that would be great. That's basically exactly what's happening - the browser entity contains / emits telemetry about a related entity, the Video Player. Being able to differentiate and relate those two entities will be very useful for our Media customers! What more information can I provide for the use case? I'm happy to get on a call to talk through it any time if that would be helpful too 😊

cjmccarthy:
Aside: We emit very similar telemetry from Mobile and Roku devices. We also want to synthesize those (separate PR coming soon - the event types are distinct so we shouldn't run into the same issues). And longer term, we are also potentially looking at synthesizing Ad Slots/Placements from the browser as entities as well. At what point might it make sense to create a Media domain or something similar to encapsulate these entity types into their own namespace?

srvaroa:
@cjmccarthy yeah there is no strong reason for using EXT- for all synthesized entities.  If we have a category that is worth creating a specific domain it's perfectly doable.

Looking forward to those PRs, we're actively discussing on the topic from the relationships synthesis perspective so those use cases are very valuable. (cc @fprats )

srvaroa:
Hi @cjmccarthy some additional questions:

> The events are all sent via PageAction from our Browser API. Because of this, they all contain the Browser entityGuid, but we would like to synthesize the Video Player as a unique and separate entity (which would ideally eventually have a synthesized relationship to its originating Browser).

Would it be feasible to *not* include the browser entityGuid in these events? If the telemetry brings no guid, and we can match the Video Player rule against it, this would result in the telemetry being stamped with the VP entity guid, and a VP entity created (usual scenario). 

To build a connection between that VP entity and the original browser entity something we could work with is if you moved the browser entityGuid to a separate attribute (e.g. a "browser.entity.guid").  This would allow us to generate a relationship between both.

entityBot (approved):

cjmccarthy:
@srvaroa Unfortunately, I don't think we can omit or rename the entityGuid, I don't think there's a way to do it with the agent API. I'll have to go test and see if I can do anything "creative" to get it to work 😜

cianBuckley:
@cjmccarthy I'm gonna go ahead and close this due to inactivity, if you feel it's still worth discussing then feel free to re-open! thanks

PR #245, opened 2021-06-17
Revert ext-host rule and golden metric changes 

IreneMLC:
### Relevant information

This reverts commit `f6362152360bf2ad0d3cced2e98d3a157e1232ae` and `a289bbf5bbbcc6ed1d4da6acd619d58d87d42ccc`. I'll really this last one after this PR is merged 

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-host/definition.yml
+++ b/definitions/ext-host/definition.yml
@@ -3,19 +3,12 @@ type: HOST
 synthesis:
   rules:
     #telemetry from datadog agent
-  - identifier: host.name
-    name: host.name
-    encodeIdentifierInGUID: true
-    conditions:
-    - attribute: metricName
-      prefix: datadog.system.
-    # opentelemetry host data from opentelemetry-collector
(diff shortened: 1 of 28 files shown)


entityBot (approved):

srvaroa:
cc @a-feld regarding b83eeb2ec836d668d9f59d570d9236a9f1a72c39.  As discussed in the original PR, given that this telemetry can't provide an `instrumentation.provider` (or an attribute that we can copy to that one) it's impossible to add a selector for the right golden / summary metrics to use in entities that are synthesized based on that rule.  This would result in a poor experience for users that get those entities created as they would not show any summary / golden metrics on the UI.  This concern is relevant because the rule is likely to match on a significant % of telemetry, so the number of impacted users would be high.

Glad to retake this conversation anytime so we can re-add this rule ASAP (cc @fprats)

srvaroa (approved):

PR #248, opened 2021-06-23
Roku entities

cjmccarthy:
### Relevant information

### Checklist

 explanation above.


entityBot (approved):

entityBot (approved):

entityBot (approved):

definitions/media-roku_app/definition.yml
```diff
@@ -0,0 +1,29 @@
+domain: MEDIA
```
cjmccarthy, line 1: I'm proposing a new Media domain which will, over time, contain at least the following entities:

- This Roku App
- Mobile Video Player (PR to come)
- [Web Video Players](https://github.com/newrelic-experimental/entity-synthesis-definitions/pull/241) (this PR needs updating)
- Ad Slots for Google Publisher Tags

definitions/media-roku_app/definition.yml
```diff
@@ -0,0 +1,29 @@
...
+# Represents all instances of a customer's Roku App monitored with New Relic's Roku agent.
+type: ROKU_APP
+
+synthesis:
+  identifier: appId
```
cjmccarthy, line 6: This gets assigned from Roku when an app/channel gets published, so it should be unique.

entityBot (approved):

definitions/media-roku_app/golden_metrics.yml
```diff
@@ -0,0 +1,37 @@
+appLaunches:
```
cjmccarthy, line 1: There are 6 metrics here - I was unable to find a precedent for more than 5 but I just performed a cursory look, so I'm hoping this is allowed.

From a UI perspective, I wondered about how 6 metrics may look in the UI. I think it will function well because we aren't expecting more than a handful of these Entity types in a given account (a customer is very unlikely to have more than a few Roku channels vs, say, hosts).

cjmccarthy:
It was unclear from the documentation, but I was curious how multi-page dashboards will work for an entity. I would ideally like to include one (and, in fact, I have here) to provide something very similar to a product-like experience out of the box.

definitions/media-roku_app/definition.yml
```diff
@@ -0,0 +1,29 @@
...
+
+  conditions:
+    - attribute: deviceGroup
+      value: Roku
+
+  tags:
+    newRelicAgent:
+    deviceGroup:
+    osName:
+
```
cjmccarthy, line 19: I had to remove the `compositeMetrics:` section in order to pass validation (it worked locally but not in the PR checker). Is this an expected result of some schema change? Just want to ensure the metrics will indeed be included 😊

naxhh: It's a change on our side. We are removing the need to define that section. The metrics will be picked up automatically.

definitions/media-roku_app/definition.yml
naxhh: What format has this appId? we have some limits on the length so if we are unsure if this id will respect those we should encode it.

definitions/media-roku_app/golden_metrics.yml
naxhh: We aim at 5 at most but I would 6 is fine too.
We can't give any guarantee that those 6 will be shown in all the UI's since some may cut the number lower.

definitions/media-roku_app/definition.yml
cjmccarthy: The Roku docs aren't great - but from our live data we can tell it's numeric (and not that large). I suspect it's just a counter on the Roku side that they increment for each new app, so I doubt we'll hit any length limits.

definitions/media-roku_app/golden_metrics.yml
cjmccarthy: Ok, I'm willing to take the risk. We can always decide to trim one off later (right? 😊)

srvaroa (approved):

entityBot (approved):

srvaroa:
@cjmccarthy this one is ready to merge  but still marked as draft, anything pending?

cjmccarthy:
@srvaroa Nope, my mistake, converted it. What's the testing path from here? When it's merged does it go to staging for some time first before rolling out to prod?

naxhh:
@cjmccarthy regarding multi-pages last word I heard is that we don't want multi pages on the overview. Ping @IreneMLC in case she has more concrete feedback on that.

entityBot (approved):

entityBot (approved):

cianBuckley (approved):

cianBuckley (commented):
@cjmccarthy Do we still want to go ahead with this new entity type?

cianBuckley:
@cjmccarthy I'm gonna close this due to inactivity, feel free to re-open if you wish!

PR #250, opened 2021-06-28
add definitions for pdns resources

ivancorrales:
### Relevant information

The **New Relic coreint team** is working on the Prometheus Exporters solution that will provide a new integration with PowerDNS. As a consequence of this, we need to define the entities synthesis. 

### Checklist

* [ X] I've read the guidelines and understand the acceptance criteria.
* [ X] The value of the attribute marked as `identifier` will be unique and valid. 
* [ X] I've confirmed that my entity type wasn't already defined. If it is I'm providing an
 explanation above.


entityBot (approved):

definitions/infra-powerdns-recursor/definition.yml
```diff
@@ -0,0 +1,16 @@
+domain: INFRA
+type: POWERDNS_RECURSOR
```
IreneMLC, line 2: The name of the folders should be `domain`-`type`, so `infra-powerdns-recursor` should be `infra-powerdns_recursor`, same for the other entity type.

entityBot (approved):

definitions/infra-powerdns-recursor/definition.yml
ivancorrales: Fixed, thanks!

IreneMLC (approved):

PR #254, opened 2021-07-01
feat: adding ext-nas entity for netgear readynas

thezackm:
### Relevant information

adding `ext-nas` entity type; tested on Netgear ReadyNAS device

**NOTE: This entity should end up in the `Other entities` section of the Explorer UI**

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-nas/summary_metrics.yml
+++ b/definitions/ext-nas/summary_metrics.yml
@@ -1,0 +1,33 @@
+# Tested on Netgear ReadyNAS
+ipAddress:
+  title: IP Address
+  unit: STRING
+  tag:
+    key: device_ip
+
+model:
+  title: Model
+  unit: STRING
+  query:
+    select: latest(SysDescr)
+    from: Metric
+    where: "provider = 'kentik-nas' AND SysObjectID = '.1.3.6.1.4.1.4526.100.16'"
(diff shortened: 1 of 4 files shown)


entityBot (approved):

srvaroa (approved):

definitions/ext-nas/summary_metrics.yml
```diff
@@ -0,0 +1,33 @@
...
+  title: IP Address
+  unit: STRING
+  tag:
+    key: device_ip
+
+model:
+  title: Model
+  unit: STRING
+  query:
+    select: latest(SysDescr)
```
naxhh, line 12: Wondering if this should also be a tag on the entity. Since it seems to make sense as one.

naxhh (approved):

definitions/ext-nas/summary_metrics.yml
thezackm: There are times when the SysDesc of a device can be several hundred characters in length. Without being able to predict this field, I moved away from that idea.

naxhh: wow. reasonable enough 😅

entityBot (approved):

Issue #249, opened 2021-06-24
Create Entity Synthesis Definitions for Entities (DB2 & Oracle) Created through InfraDBPluginJava


sri-shetty:
Since NR Infra (and OHI) agents doesn't support AIX platform, Customers use FIT InfraDBPluginJava (https://source.datanerd.us/FIT/InfraDBPluginJava) to monitor databases like DB2, Oracle and other DBs. These entities are not visible in the Entity Explorer or Workloads. 

I would like to create Entity Synthesis Definition for DB2 and Oracle database for Manulife.

srvaroa:
Hi @sri-shetty, thanks for raising this one. 

We can help you craft those rules, do you have some samples of telemetry we could look at?

sri-shetty:
[db2 Event Type Keyset](https://insights-embed.newrelic.com/embedded_widget/-2kjrJQk7md41GBvKWA2cW_xgAUFWNXu)
[Oracle Event Type Keyset](https://insights-embed.newrelic.com/embedded_widget/N_FwRI2tAmYF4jkdew0L8T4qY7UG2jyO)
[DB2-QueryName.csv](https://github.com/newrelic-experimental/entity-synthesis-definitions/files/6727804/DB2-QueryName.csv)
[DB2-Unique-QueryName.csv](https://github.com/newrelic-experimental/entity-synthesis-definitions/files/6727805/DB2-Unique-QueryName.csv)
[OracleMetrics.csv](https://github.com/newrelic-experimental/entity-synthesis-definitions/files/6727806/OracleMetrics.csv)

Please ping me sshetty@newrelic.com - if you need additional details.

srvaroa:
Hi @sri-shetty,  registering those types will require a PR adding the relevant configuration files for each type based on that telemetry.  From what I saw in a couple of examples (db2, oracle) this would look pretty similar to other existing types like [ext-ibmmqqueue](https://github.com/newrelic-experimental/entity-synthesis-definitions/blob/master/definitions/ext-ibmmqqueue/definition.yml).

This is a draft example of how the file would look for Oracle.  It should be added at `definitions/ext-oracle/definitions.yml`

Note that I've wild-guessed the values for `name` and `identifier` attributes.   The `identifier` should be a telemetry attribute that is unique within an account. I used `db_unique_name` but you should check whether this attribute is the right one. 

```
domain: EXT
type: ORACLE

synthesis:
  name: database
  identifier: db_unique_name
  conditions:
[...]

PR #256, opened 2021-07-05
Db2 and Oracle instances

naxhh:
### Relevant information

Add synth definitions for DB2 and Oracle instances.

I went with plain DB2 & Oracle for the types. I notice EXT-REDIS is just redis while infra is REDIS_INSTANCE. So let me know if you prefer one or the other.

Notice these are not "databases" but "instances". One instance can contain multiple databases.

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-db2/definition.yml
+++ b/definitions/ext-db2/definition.yml
@@ -1,0 +1,17 @@
+domain: EXT
+type: DB2
+
+synthesis:
+  identifier: instance
+  name: instance
+  encodeIdentifierInGUID: true
+  conditions:
+  - attribute: event_type
+    value: db2
+  tags:
+    database:
+    hostname:
+
+configuration:
+  entityExpirationTime: EIGHT_DAYS
+  alertable: true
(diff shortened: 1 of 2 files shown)


entityBot (approved):

naxhh:
relates #249 

@sri-shetty this is a proposal based on the data you provided to get the ball rolling.

It still requires dashboards and golden-metrics and summary-metrics. I understand you will provide those.


For our side: We will need to create the JS widgets + I think the alertable field still requires manual update on unify?
Finally this should be under "On host" ui category.

definitions/ext-db2/definition.yml
```diff
@@ -0,0 +1,17 @@
+domain: EXT
```
IreneMLC, line 1: A small comment with a description about the type might be useful for the future

IreneMLC (approved):

definitions/ext-db2/definition.yml
naxhh: Any suggestion? Everything that I can come up with seems not really useful (like: IMB DB2 databases instances)

IreneMLC: I was thinking of the notice that you added to the PR description actually, about these entities being instances/

naxhh:
closing @sri-shetty is working in a PR with GM & dashboards + the defs.

PR #272, opened 2021-07-16
Make containerId single value.

srvaroa:
There are some containers that use the same name (which determines the identifier) but get restarted often.   As the tag is set to multi value, we'll keep accumulating the containerId tag.

### Relevant information

### Checklist

 explanation above.


entityBot (approved):

gvassallo (approved):
ship it!

srvaroa:
Got confirmation from @douglascamata that the multi value is not known to be used.

PR #278, opened 2021-07-21
Reintroduce oracle & db2

naxhh:
We reverted oracle & db2 changes. Keep this around to make the required changes after talking with core-integrations


entityBot (dismissed):

definitions/ext-db2/definition.yml
```diff
@@ -0,0 +1,17 @@
+domain: EXT
+type: DB2
+synthesis:
+  name: database
```
naxhh, line 4: One doubt is if this should be instance.

Since other db's seem to use instance not database name.

definitions/ext-oracle/definition.yml
```diff
@@ -0,0 +1,17 @@
+domain: EXT
```
naxhh, line 1: We notice an INFRA-ORACLEINSTANCE type already exists. So this most likley needs to be merged into that type.

IreneMLC:
@entityBot

entityBot (changes requested):
beep boop bop.

 ```
* Error during VALIDATING: sum(1), sum(5), sum( - Exception: Retrieving more than one value from the Select clause is not allowed.

* Error during VALIDATING: sum(, sum( - sum( - Exception: Retrieving more than one value from the Select clause is not allowed.
```

entityBot (approved):

sri-shetty:
Any progress with pushing db2 definition to staging/production? What is the plan on merging Oracle OHI and Extension definitions. Were you come to a conclusion on whether to merge to keep them separate?

IreneMLC:
Hi @sri-shetty we have a ticket in our sprint to follow up on this and reach a conclusion, we'll keep you updated. If this is blocking you please let us know when you'd need this done so we can work with that in mind. Thanks in advance!

sri-shetty:
Thanks @IreneMLC - It seems there is no conflict with the db2 definition. It would be great if you could push db2 definition to production next week.

definitions/ext-db2/definition.yml
IreneMLC: Hi @sri-shetty with the db2 definition the issue is what Ignacio mentioned here, could you share a sample of the telemetry for us to check it? is the 'database' attribute going to be unique between other instances of this type?

sri-shetty: @IreneMLC - database attribute in the db2 event type is unique. I have uploaded the sample events for db2 into staging. Here is the permalink to staging. [Staging db2 event type](https://staging-one.newrelic.com/-/0nVjYAKYlj0). I'll email you the json containing db2 events.

IreneMLC: Thanks for the information, since you have the domain knowledge we can proceed with this. If you detect any issues you can modify the attributes used to synthesize the entity, the only consequence of that is that the GUID will change if the identifier is changed.

entityBot (approved):

entityBot (approved):

IreneMLC:
I've removed the `ext-oracle` definition from this PR so we can merge `ext-db2`

IreneMLC (approved):

PR #292, opened 2021-08-24
feat: adding kentik-default entity type

thezackm:
### Relevant information

adding a default entity type for telemetry received from Kentik's ktranslate agent

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-kentik_default/dashboard.json
+++ b/definitions/ext-kentik_default/dashboard.json
@@ -1,0 +1,51 @@
+{
+	"name": "Kentik-Default",
+	"description": null,
+	"pages": [
+		{
+			"name": "Kentik-Default",
+			"description": null,
+			"widgets": [
+				{
+					"visualization": {
+						"id": "viz.markdown"
+					},
+					"layout": {
+						"column": 1,
+						"row": 1,
+						"height": 4,
+						"width": 10
+					},
+					"title": "",
+					"rawConfiguration": {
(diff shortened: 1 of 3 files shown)


entityBot (approved):

entityBot (approved):

entityBot (approved):

naxhh:
@entityBot

entityBot (approved):

naxhh:
validation fails because one widget is missing `nrqlQueries` that is probably an oversight on our side since so far we only had nrql widgets. I'll look into that.

srvaroa:
Hi @thezackm, this looks OK, I just had a couple of questions:

- Would it make more sense to call these `kentik-device` or similar? (I understand they represent a Kentik device).  I see the `provider` is set to `kentik-default` too so it may be that this is the correct terminology anyway, let me know if that's the case.

- If I understand correctly, the intent is that this rule will capture Kentik telemetry that can't be  matched to one of the other specific devices for which we have entities. So what you do is capture as a generic Kentik entity, and the plan is that the user would at some point tweak their integration so that the telemetry starts matching another rule (for EXT-ROUTER, etc.)

This has an implication I want to make sure you're aware of.  When the user sends telemetry that matches on this rule (phase 1) we'll stamp the telemetry with the corresponding EXT-KENTIK_DEFAULT GUID.  If later (phase 2) the user completes the integration and telemetry matches EXT-ROUTER for example, then the telemetry will have a completely different GUID.  That is, there will be two different entities (one for phase 1, another for phase 2).  The dashboards for the first will show data from phase 1, but not later.  The dashboards for the second will show data from phase 2, but not earlier.  Also, anything that relied on the first GUID will not be updated to use the second. E.g. if you have alert conditions targetting the first GUID those will not get updated to use the second GUID. Does that make sense? Is it ok for you?

Side note: we have a bug in the validator, @naxhh is fixing that as we speak.

entityBot (approved):

jfjoly:
> * Would it make more sense to call these `kentik-device` or similar? (I understand they represent a Kentik device).  I see the `provider` is set to `kentik-default` too so it may be that this is the correct terminology anyway, let me know if that's the case.

It represents a device that is being polled by Kentik ktranslate agent. That is different from a Kentik device.
Maybe a more generic name would be snmp-default although I would keep the proposal of Zack.

jfjoly:
> * If I understand correctly, the intent is that this rule will capture Kentik telemetry that can't be  matched to one of the other specific devices for which we have entities. So what you do is capture as a generic Kentik entity, and the plan is that the user would at some point tweak their integration so that the telemetry starts matching another rule (for EXT-ROUTER, etc.)

Correct, with the same comment as above. This is not a Kentik device but a device being polled by the Kentik agent.

srvaroa:
Thanks @jfjoly ! We'll merge this now.

For the record: the CLA check is failing but Zach has signed it previously, and is also a NR employee.

jfjoly:
> When the user sends telemetry that matches on this rule (phase 1) we'll stamp the telemetry with the corresponding EXT-KENTIK_DEFAULT GUID. If later (phase 2) the user completes the integration and telemetry matches EXT-ROUTER for example, then the telemetry will have a completely different GUID. That is, there will be two different entities

I think this is acceptable.
To confirm, the phase 1 entity will disappear after 4 hours (expiration time)?

srvaroa:
@jfjoly yup

thezackm:
Thanks for the comments @srvaroa !

1) I think "default" is the better term since it aligns with the code from the vendor. 

2) that is 100% understood and expected. I put a 4 hour expiration on these specifically to help with the potential overlap and we don't want or expect any users to keep these around long enough to have critical dashboards/alerts. We really want to guide them towards building "real" entities for that.

PR #306, opened 2021-09-05
Add new mlops entities

neta-signifai:
### Relevant information

Add two new types of entities to the MLOPS domain (endpoint and ml_training) with synthesis rules 

### Checklist

* [ V] I've read the guidelines and understand the acceptance criteria.
* [ ]V The value of the attribute marked as `identifier` will be unique and valid. 
* [ V] I've confirmed that my entity type wasn't already defined. If it is I'm providing an
 explanation above.


entityBot (approved):

entityBot (commented):
LGTM waiting to see some data to verify the implementation

definitions/mlops-machine_learning_training/definition.yml
```diff
@@ -0,0 +1,19 @@
+domain: MLOPS
+type: MACHINE_LEARNING_TRAINING
+synthesis:
+  rules:
+    - identifier: aws..aws.sagemaker.trainingjobs.TrainingJobName
```
naxhh, line 5: mostly a doubt but is the job name unique? or we should use the ARN? (for both entities)

naxhh (approved):

definitions/mlops-machine_learning_training/definition.yml
naxhh: discussed internally

entityBot (approved):

definitions/mlops-endpoint/definition.yml
```diff
@@ -0,0 +1,22 @@
+domain: MLOPS
+type: ENDPOINT
```
srvaroa, line 2: Are endpoints a Sagemaker-specific concept? I'm wondering if we should use SM_ENDPOINT or something like that.

We have another ENDPOINT type in the EXT- domain.  This will not cause problems anywhere, but for historical reasons telemetry doesn't get stamped with the domain, so this case will generate ambiguities when we process them.  It won't break anything, but I'm wondering if it's worth trying to avoid it anyway.

definitions/mlops-machine_learning_training/definition.yml
```diff
@@ -0,0 +1,19 @@
+domain: MLOPS
+type: MACHINE_LEARNING_TRAINING
```
srvaroa, line 2: Would TRAINING_JOB be more accurate?

definitions/mlops-endpoint/definition.yml
neta-signifai: changed it to MACHINE_LEARNING_ENDPOINT (we would want to include endpoints from other sources in the future)

definitions/mlops-machine_learning_training/definition.yml
neta-signifai: I want to leave it more general (although might change in the future)

naxhh: Changing it in the future will mean breaking all entities and creating new ones.
With their alerts, metrics, favorites etc.

If we think in the future should be more specific better do it now.

entityBot (approved):

definitions/mlops-machine_learning_training/definition.yml
neta-signifai: ok, changed, thanks

entityBot (approved):

srvaroa (approved):

PR #308, opened 2021-09-09
adding Istio Service Entity

ericmittelhammer:
### Relevant information

Adding definition, golden metrics, summary metrics, and summary dashboard for an **Istio Service** entity

### Checklist

 explanation above.

cc: @thezackm

### DIFF
--- a/definitions/ext-istio_service/definition.yml
+++ b/definitions/ext-istio_service/definition.yml
@@ -1,0 +1,47 @@
+domain: EXT
+type: ISTIO_SERVICE
+synthesis:
+  rules:
+  - identifier: "label.service.istio.io/canonical-name"
+    name: "label.service.istio.io/canonical-name"
+    conditions:
+    - attribute: metricName
+      prefix: istio_request
+
+  - identifier: "label.service.istio.io/canonical-name"
+    name: "label.service.istio.io/canonical-name"
+    conditions:
+    - attribute: metricName
+      prefix: istio_response
+
+  - identifier: "label.service.istio.io/canonical-name"
+    name: "label.service.istio.io/canonical-name"
+    conditions:
+    - attribute: metricName
+      prefix: istio_tcp
+
+  tags:
+    label.service.istio.io/canonical-name:
+    label.service.istio.io/canonical-revision:
+    label.app:
+    label.version:
+    label.istio:
+    label.istio.io/rev:
+    label.operator.istio.io/component:
+    k8s.cluster.name:
+    clusterName:
+    namespaceName:
+    deploymentName:
+
+goldenTags:
+  - clusterName
+  - namespaceName
+  - label.app
+
+dashboardTemplates:
+  newRelic:
+    template: dashboard.json
+
+configuration:
+  entityExpirationTime: FOUR_HOURS
+  alertable: true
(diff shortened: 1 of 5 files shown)


definitions/ext-istio_service/golden_metrics.yml
```diff
@@ -0,0 +1,16 @@
+throughput:
+  title: Throughput
+  query:
+    select: rate(sum(istio_requests_total), 1 SECOND) WHERE reporter = 'destination'
+    from: Metric
+errorRate:
+  title: Error rate
+  query:
+    select: (filter(rate(sum(istio_requests_total), 1 minute), WHERE response_code LIKE '5%') / rate(sum(istio_requests_total), 1 minute)) * 100 WHERE reporter = 'destination'
```
thezackm, line 9: curious how the golden metric harvester would handle this query

definitions/ext-istio_service/summary_metrics.yml
```diff
@@ -0,0 +1,31 @@
...
+    eventId: entity.guid
+errorRate:
+  title: Error rate
+  unit: PERCENTAGE
+  query:
+    select: (filter(rate(sum(istio_requests_total), 1 minute), WHERE response_code LIKE '5%') / rate(sum(istio_requests_total), 1 minute)) * 100 WHERE reporter = 'destination'
+    from: Metric
+    eventId: entity.guid
+responseTimeMs:
+  title: Response time (ms)
```
thezackm, line 26: should this be `Response time` (without the `ms` designation)?

ericmittelhammer: Yep.  Good catch.  Had copied over from the golden metrics (where I followed the convention in [ext-service](https://github.com/newrelic/entity-definitions/blob/a21373c72a60b465c8ec01ad226a90506201a21d/definitions/ext-service/golden_metrics.yml#L33) ).
Will fix

definitions/ext-istio_service/golden_metrics.yml
ericmittelhammer: Because of the filter?  Looks like it's used elsewhere:  https://github.com/newrelic/entity-definitions/search?q=filter+filename%3Agolden_metrics.yml

thezackm: I retract my concern then 😄

definitions/ext-istio_service/definition.yml
```diff
@@ -0,0 +1,47 @@
+domain: EXT
+type: ISTIO_SERVICE
+synthesis:
+  rules:
+  - identifier: "label.service.istio.io/canonical-name"
```
srvaroa, line 5: How long can this field be? we probably want to set encoding to true

ericmittelhammer: @srvaroa I don't know the maximum limit, so to be safe, I've added `encodeIdentifierInGUID: true` to each of the rules

srvaroa: Thanks!

definitions/ext-istio_service/golden_metrics.yml
srvaroa: @rcrocker can you confirm that these are ok? (I know we may have a couple of problematic queries still in the definitions so just to be on the safe side)

naxhh (approved):

srvaroa (approved):

PR #314, opened 2021-09-14
Add Infra processes as entities

khpeet:
### Relevant information
Added Infrastructure agent reported processes as entities

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-processes/definition.yml
+++ b/definitions/infra-processes/definition.yml
@@ -1,0 +1,32 @@
+domain: INFRA
+type: PROCESS
+
+synthesis:
+  name: commandName
+  identifier: commandLine
+  encodeIdentifierInGUID: true
+  conditions:
+  - attribute: eventType
+    value: ProcessSample
+
+  tags:
+    hostname:
+      multiValue: true
+    app_support_team:
+      multiValue: true
+    datacenter:
+      multiValue: true
+    environment:
+      multiValue: true
+    userName:
+      multiValue: true
+    processDisplayName:
+      multiValue: true
+
+dashboardTemplates:
+  newRelic:
+    template: dashboard.json
+
+configuration:
+  entityExpirationTime: DAILY
+  alertable: true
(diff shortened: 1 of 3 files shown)


entityBot (approved):

khpeet:
@srvaroa - Added this per our discussion to test it out

srvaroa:
Thanks @khpeet !

Do you have a rough estimate of how many entities we expect to generate with this rule? (I assume it's quite large).  Asking mainly in case it's worth finding a condition that could narrow to specific process types (e.g. `commandName="java"`)

khpeet:
@srvaroa - I don't have an estimate, but I expect it will be huge number of entities. Is it possible to test without a filter in staging to determine if we need a filter? I'm also not sure why the dashboard.json check failed here since I ran it through the dashboard sanitizer.

srvaroa:
Hi @khpeet I've checked and our internal account does have a significant number.  It'd be worth starting with a more restrictive rule to do a more controlled rollout so I'm trying to think of a way to narrow them down.  Is there a particular type of process you're more interested in? (e.g. we can add a rule that matches on `commandName=java` for example and that will fetch a smaller portion.  We can widen the filter gradually after that.

definitions/infra-processes/definition.yml
```diff
@@ -0,0 +1,32 @@
...
+      multiValue: true
+    processDisplayName:
+      multiValue: true
+
+dashboardTemplates:
+  newRelic:
+    template: dashboard.json
+
+configuration:
+  entityExpirationTime: DAILY
```
srvaroa, line 31: Have you considered shorter TTLs? (e.g. I assume in the general case a process that is restarting does so in shorter timespans, so it may be enough to keep the entity alive for 4h)

entityBot (approved):

entityBot (approved):

entityBot (approved):

naxhh:
@khpeet  @srvaroa where this ended up? should i close it?

khpeet:
@naxhh - this can be closed. We found that this isn't achievable in the current state since an eventType can only have 1 entity guid.

PR #326, opened 2021-09-23
Add synthesis definition for opentelemetry-collector host metrics.

jlegoff:
### Relevant information

This is a test in order to check whether opentelemetry hosts can be synthesized as infra/host entities. I added a "instrumentation.provider" which is currently not sent by the collector and will prevent entities from being synthesized in practice. However, it allows to manually tune the collector to test the entity synthesis.

I also added a golden metric to check whether we can get the same values as the infra agent using the `derivative` function.

### Checklist

 explanation above.


entityBot (changes requested):
beep boop bop.

 ```
* Error PARSING domainType: 'infra-host' on BaseEntityDefinition - ValueInstantiationException: Cannot construct instance of `EntitySynthesisRule`, problem: must not contain more than one condition
 at [Source: (StringReader); line: 46, column: 1] (through reference chain: BaseEntityDefinition["synthesis"]->CustomEntitySynthesisRules["rules"]->ArrayList[0])
```

File: definitions/infra-host/definition.yml [hunk 1]
```diff
@@ -13,6 +13,36 @@ goldenTags:
...
+# opentelemetry semantic conventions for cloud instances
+- cloud.provider
+- cloud.account.id
+- cloud.region
+- cloud.availability_zone
+- cloud.platform
+synthesis:
+  rules:
+    # opentelemetry host data from opentelemetry-collector
+    - identifier: host.id
```
cianBuckley, line 25: infra|na hosts use a privately generated domainId, right? Is this attr value the same?

cianBuckley: ah, i think i get what's going on - this is telemetry from the nr agent, but you're somehow flagging it with additional instrumentation.provider info, right?

jlegoff: this is telemetry coming from the opentelemetry collector. I'm trying to investigate we can classify these as infra hosts (as opposed to ext-hosts) in order to view all your hosts together, regardless of their source.

entityBot (approved):

cianBuckley (approved):

entityBot (approved):

PR #346, opened 2021-10-15
Golden Metrics - adding units 1

stephelhajj:
### Relevant information

Added units to golden-metrics.yml files 

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-awsdocdbinstance/golden_metrics.yml
+++ b/definitions/infra-awsdocdbinstance/golden_metrics.yml
@@ -1,5 +1,6 @@
 cpuUsage:
   title: CPU usage
+  unit: PERCENTAGE
   queries:
     aws:
       select: average(aws.docdb.CPUUtilization.byInstance)
@@ -14,6 +15,7 @@ cpuUsage:
       eventName: entityName
 databaseConnections:
   title: Database connections
+  unit: COUNT
   queries:
     aws:
       select: average(aws.docdb.DatabaseConnections.byInstance)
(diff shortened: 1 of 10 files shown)


definitions/infra-awsalbtargetgroup/golden_metrics.yml
```diff
@@ -28,6 +30,7 @@ unhealthyHosts:
       eventName: entityName
 responseTime:
   title: Average response time
+  unit: SECONDS
   queries:
     aws:
       select: average(aws.applicationelb.TargetResponseTime)
```
naxhh, line 36: double checking: is this metric actually on seconds or we need to convert it to seconds?

definitions/infra-awsapigatewayapi/golden_metrics.yml
```diff
@@ -28,6 +30,7 @@ errors4XxAnd5Xx:
       eventName: entityName
 latencyMs:
   title: Latency (ms)
+  unit: SECONDS
```
naxhh, line 33: from the title seems this is on miliseconds right? so we should convert it into seconds

definitions/infra-awsapigatewayresourcewithmetrics/golden_metrics.yml
```diff
@@ -28,6 +30,7 @@ errors4XxAnd5Xx:
       eventName: entityName
 latencyMs:
   title: Latency (ms)
+  unit: SECONDS
```
naxhh, line 33: same here

PR #379, opened 2021-11-18
feat: add flex host synthesis

Kav91:
### Relevant information

Synthesis for Flex host generated data. 
This is a requirement for the AWS/MAP partner project.

### Checklist

 explanation above.


entityBot (approved):

entityBot (approved):

definitions/ext-flex_host/definition.yml
```diff
@@ -0,0 +1,30 @@
+domain: EXT
+type: FLEX_HOST
```
naxhh, line 2: wondering if this should be under `EXT-HOST` instead of it's own type.

Thoughts @srvaroa

definitions/ext-flex_host/definition.yml
```diff
@@ -0,0 +1,30 @@
...
+synthesis:
+  name: displayName
+  identifier: displayName
```
naxhh, line 5: Is there any limits on the displayName attribute? if not we may want to add `      encodeIdentifierInGUID: true`

definitions/ext-flex_host/definition.yml
```diff
@@ -0,0 +1,30 @@
...
+  conditions:
+  - attribute: eventType
+    value: FlexHostSample
+  tags:
+    arch:
+      multiValue: false
+    coreCount:
+      multiValue: false
+    displayName:
```
naxhh, line 14: this doesn't need to be a tag. Is already the name of the entity, and you can filter by name if required.

definitions/ext-flex_host/definition.yml
```diff
@@ -0,0 +1,30 @@
...
+      multiValue: false
+    fullHostname:
+      multiValue: false
+    longHostname:
```
naxhh, line 18: wondering on the difference between fullHostname, longHostname & shortHostname do we really need them all?

definitions/ext-flex_host/summary_metrics.yml
```diff
@@ -0,0 +1,36 @@
+displayName:
```
naxhh, line 1: the name of the entity will be there without needing to add it here so I would remove this one.

entityBot (approved):

definitions/ext-flex_host/definition.yml
Kav91: image
Have removed one of the hostname attributes, but it is to keep parity with how the infra agent does it.

definitions/ext-flex_host/definition.yml
Kav91: removed thanks

definitions/ext-flex_host/summary_metrics.yml
Kav91: removed thanks

definitions/ext-flex_host/definition.yml
Kav91: no limits, have added this now thx

definitions/ext-flex_host/definition.yml
Kav91: The metrics and dashboards included are very different, would this cause any issues? The current ext-host has DD based metrics and it looks quite limited in comparison.

Also if we were to do this, is there a good example I can use as a reference? thx

entityBot (approved):

definitions/ext-flex_host/definition.yml
```diff
@@ -0,0 +1,28 @@
...
+    localIp: 
+      multiValue: false
+    shortHostname: 
+      multiValue: false
+      entityTagName: hostname
+dashboardTemplates:
+  newRelic:
+    template: dashboard.json
+configuration:
+  alertable: true
```
naxhh, line 28: is the default ttl good enough?

entityBot (approved):

entityBot (approved):

definitions/ext-flex_host/definition.yml
Kav91: @naxhh have set it to QUARTERLY now, this should be sufficient.

definitions/ext-flex_host/definition.yml
naxhh: Yeah shape of data is not a problem in this case, we can provide different representations.

Yo have an example here: https://github.com/newrelic/entity-definitions/blob/main/definitions/ext-service/definition.yml


This is my main question to move this PR forward so let me ping the team to have more input on how we should do it

entityBot (approved):

entityBot (approved):

Kav91:
Hi @naxhh have re-implemented this under EXT-HOST can you review it again please and let me know if any issues? Thanks

definitions/ext-deeper_connect/definition.yml
```diff
@@ -30,4 +30,5 @@ dashboardTemplates:
   newRelic:
     template: dashboard.json
 configuration:
-  alertable: true
\ No newline at end of file
+  alertable: true
+  entityExpirationTime: QUARTERLY
```
naxhh, line 34: how is this related to flex-host ?

definitions/ext-host/definition.yml
```diff
@@ -10,9 +10,20 @@ synthesis:
...
+      name: displayName
+      encodeIdentifierInGUID: true
+      conditions:
+      - attribute: eventType
+        value: FlexSystemSample
+
 dashboardTemplates:
-  newRelic:
-    template: dashboard.json
+  flexagent:
```
naxhh, line 22: This value needs to exist as the `intrumentation.provider` attribute of the telemetry. would that happen?

I was searching for some data but didn't find any I could use as reference

definitions/ext-host/definition.yml
```diff
@@ -10,9 +10,20 @@ synthesis:
...
+    template: flex-dashboard.json
+  datadog:
+    template: datadog-dashboard.json
+
 configuration:
-  entityExpirationTime: DAILY
+  entityExpirationTime: QUARTERLY
```
naxhh, line 28: Im leaning on keeping it as daily. that's also the default in INFRA-HOST: https://github.com/newrelic/entity-definitions/blob/main/definitions/infra-host/definition.yml#L46

so making every host the same expiration makes sense to me and is easier on future merges.

what do you think?

definitions/ext-host/golden_metrics.yml
```diff
@@ -1,9 +1,47 @@
...
+    flexagent:
+      select: latest(loadAverageFifteenMinute)
+      from: FlexSystemSample
+      eventId: entity.guid
+
+cpuPercent:
+  title: CPU Percent
+  unit: PERCENTAGE
+  queries:
+    flexagent:
```
naxhh, line 15: we must provide a query for each provider.

definitions/ext-host/summary_metrics.yml
```diff
@@ -0,0 +1,36 @@
...
+  tag:
+    key: shortHostname
+  title: hostname
+  unit: STRING
+
+cpuPercent:
+  title: CPU Percent
+  unit: PERCENTAGE
+  queries:
+    flexagent:
```
naxhh, line 17: same here. a query for all sources should be provided

definitions/ext-host/definition.yml
Kav91: Needs to be QUARTERLY for this scenario: User runs the Flex agent temporarily to harvest data for a period of time, it will not run indefinitely as during migrations they just need to capture for a short period to understand the environment and entities within it.
Migrations can take weeks to months, so the entity needs to live far longer then a day.

definitions/ext-deeper_connect/definition.yml
Kav91: Was accidental, it can be quarterly, but yes it is unrelated to the PR in question, I will remove it on the next commit.

definitions/ext-host/definition.yml
naxhh: hmm that makes sense. i was wondering if this should go under infra-host but that seems a rather specific case good enough to be on ext-host

definitions/ext-host/golden_metrics.yml
naxhh: A host is a host in itself. The golden metrics of a host are always the same regardless of how they are queried internally.

In this case you have provided the queries for the existing golden metrics just fine.
Since you added a new one we need to also add the query for the existing sources, in this case datadog.

As far as I could find this is the most likely candidate `max(datadog.system.cpu.system)`.
And if someone complains we are happy to change the query so please add it for datadog.

definitions/ext-deeper_connect/definition.yml
naxhh: Yes please. let's keep this flex related and if you need to we can do a new PR with that change

entityBot (approved):

entityBot (approved):

definitions/ext-deeper_connect/definition.yml
Kav91: reverted back to original

entityBot (approved):

definitions/ext-host/summary_metrics.yml
Kav91: added

definitions/ext-host/golden_metrics.yml
Kav91: added the dd queries

entityBot (approved):

definitions/ext-host/definition.yml
Kav91: Yes, I've updated the integration to include it.

image

account id: 1061074

definitions/ext-host/definition.yml
```diff
@@ -10,9 +10,20 @@ synthesis:
...
+      conditions:
+      - attribute: eventType
+        value: FlexSystemSample
+
 dashboardTemplates:
-  newRelic:
-    template: dashboard.json
+  flexagent:
+    template: flex-dashboard.json
+  datadog:
```
Kav91, line 24: @naxhh will we have issues here given the datadog telemetry may not have the instrumentation.provider attribute?

naxhh: I was checking the data and you are right indeed. Datadog agent doesn't provide a inst.provider.

Put it first in all rules so it will default to that one when the instrumentation.provider doesn't exist

definitions/ext-host/definition.yml
naxhh: perfect thanks!

entityBot (approved):

definitions/ext-host/definition.yml
Kav91: @naxhh have moved it first, would it be okay to merge now?

entityBot (approved):

cianBuckley (approved):

apoloa (approved):

entityBot (approved):

PR #387, opened 2021-12-02
change alertable to false

rodolfocal:
### Relevant information

Changing `alertable` from `true` to `false` in EXT-ENDPOINT, which represents a network endpoint.
Changing `alertable` from `true` to `false` on Infra entities which should not have `health`.

### Checklist

 explanation above.


entityBot (approved):

migruealt:
Looks good to me

migruealt (approved):
👍  thanks!

hredondo (approved):

PR #398, opened 2021-12-30
INFRA-HOST: add cpu gm for otel

rubenruizdegauna:
### Relevant information

This is adding the cpu golden metric definition for hosts monitored by the opentelemetry collector.

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-host/golden_metrics.yml
+++ b/definitions/infra-host/golden_metrics.yml
@@ -12,6 +12,12 @@ cpuUsage:
       from: SystemSample
       eventId: entityGuid
       eventName: entityName
+    opentelemetry:
+      select: average(1 - system.cpu.utilization)
+      where: state = 'idle'
+      from: Metric
+      eventId: entity.guid
+      eventName: entity.name
 memoryUsage:
   title: Memory usage (%)
   unit: PERCENTAGE
--- a/definitions/infra-host/golden_metrics.yml
+++ b/definitions/infra-host/golden_metrics.yml
@@ -13,7 +13,7 @@ cpuUsage:
       eventId: entityGuid
       eventName: entityName
     opentelemetry:
-      select: average(1 - system.cpu.utilization)
+      select: average(1 - system.cpu.utilization) * 100
       where: state = 'idle'
       from: Metric
       eventId: entity.guid


entityBot (approved):

danipl (approved):

definitions/infra-host/golden_metrics.yml
```diff
@@ -12,6 +12,12 @@ cpuUsage:
       from: SystemSample
       eventId: entityGuid
       eventName: entityName
+    opentelemetry:
+      select: average(1 - system.cpu.utilization)
```
otaviocarvalho, line 16: I was able to find other examples where the average is used in a calculation [[1](https://github.com/newrelic/entity-definitions/blob/2a5cb3d03bbd442037d60e72f355e01318a9d908/definitions/infra-gcphttploadbalancer/golden_metrics.yml#L20),[2](https://github.com/newrelic/entity-definitions/blob/c83c8348131e5980a2bc6292e18b2c33c9bb6618/definitions/apm-application/golden_metrics.yml#L4)]. However, I was not able to find any other case where a calculation is present inside of the average operator.

Are we sure harvester will be able to parse this one correctly? cc @naxhh @danipl

otaviocarvalho (approved):

definitions/infra-host/golden_metrics.yml
danipl: Hi @otaviocarvalho!!, at this time we're applying the same validations that harvester does over these queries.

apoloa (approved):

definitions/infra-host/golden_metrics.yml
rubenruizdegauna: This would have the same effect as `1 - average(system.cpu.utilization)` . Would it better to change it or is it ok as it is?

definitions/infra-host/golden_metrics.yml
```diff
@@ -12,6 +12,12 @@ cpuUsage:
       from: SystemSample
       eventId: entityGuid
       eventName: entityName
+    opentelemetry:
```
naxhh, line 15: This requires the telemetry to bring an `instrumentation.provider=opentelemtry` attribute. Is that actually happening?

Last time I talked with Julien I understood otel didn't bring this

definitions/infra-host/golden_metrics.yml
naxhh: As Dani noted, if the bot is not complaining this should work as it is, we can double check in staging after merging

jlegoff: @rubenruizdegauna you should multiply this result by 100, otherwise this won't give a percentage

jlegoff (changes requested):

definitions/infra-host/golden_metrics.yml
rubenruizdegauna: Thanks @jlegoff , I added it in a different commit : https://github.com/newrelic/entity-definitions/pull/398/commits/af5863c6458616b5dcaf05193c07cfa2e8d94658

definitions/infra-host/golden_metrics.yml
naxhh: yes. that one. nice :D

naxhh (approved):

entityBot (approved):

PR #411, opened 2022-01-24
Add new entity type ext-sap_system

edyu002:
### Relevant information

### Checklist

* [ x] I've read the guidelines and understand the acceptance criteria.
 explanation above.

### DIFF
--- a/definitions/ext-sap_system/definition.yml
+++ b/definitions/ext-sap_system/definition.yml
@@ -1,0 +1,24 @@
+domain: EXT
+type: SAP_SYSTEM
+goldenTags:
+  - instrumentation.provider
+
+synthesis:
+  rules:
+  # Telemetry from New Relic SAP integration
+  - identifier: SYS_ID
+    name: SYS_ID
+    encodeIdentifierInGUID: true
+    conditions:
+    - attribute: SAP_ETYPE
+      value: SYSTEM
+    tags:
+      instrumentation.provider:
+
+dashboardTemplates:
+  SAP:
+    template: dashboard.json
+
+configuration:
+  entityExpirationTime: QUARTERLY
+  alertable: true
--- a/definitions/ext-sap_system/definition.yml
+++ b/definitions/ext-sap_system/definition.yml
@@ -2,7 +2,7 @@ domain: EXT
 type: SAP_SYSTEM
 goldenTags:
   - instrumentation.provider
-  - sap.funtional_area
+  - sap.functional_area
   - sap.system
 
 synthesis:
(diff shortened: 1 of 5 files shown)


entityBot (approved):

edyu002:
This pull request is to add ext-sap_system entity type by the SAP observability team. This entity type defines an SAP system.

definitions/ext-sap_system/definition.yml
```diff
@@ -0,0 +1,24 @@
...
+  rules:
+  # Telemetry from New Relic SAP integration
+  - identifier: SYS_ID
+    name: SYS_ID
+    encodeIdentifierInGUID: true
+    conditions:
+    - attribute: SAP_ETYPE
+      value: SYSTEM
+    tags:
+      instrumentation.provider:
```
naxhh, line 16: You can remove this, instrumentation.provider will be always added as a tag if the attribute is present.

naxhh (approved):
LGTM minor thing to cleanup

definitions/ext-sap_system/definition.yml
edyu002: Ok. Will remove it.

entityBot (approved):

danipl (approved):

entityBot (approved):

definitions/ext-sap_system/definition.yml
```diff
@@ -0,0 +1,30 @@
...
+  # Telemetry from New Relic SAP integration
+  - identifier: SAP_EID
+    name: SYS_ID
+    encodeIdentifierInGUID: true
+    conditions:
+    - attribute: SAP_ETYPE
+      value: SYSTEM
+    tags:
+      sap.functional_area:
+      SYS_ID:
```
naxhh, line 19: this is the name of the entity, why it needs to be a tag?

Sorry i just noticed this now

edyu002: The intention for adding SYS_ID attribute as the sap.system is to use it for all applicable entity types that belong to a system, including the system itself; then use the tag to create a workload that group different type of entities of a sap systems. Then this will allow us to present a view from SAP System organized. If there is a way of taking advantage of this attribute being the name of the sap_system entity itself, we'd like to know and consider to use it.

entityBot (approved):

definitions/ext-sap_system/definition.yml
naxhh: with this you are looking at a query like: `tags.sap_system = 'system name'` without this you would need to do `tags.sap_system = 'system name' and name = 'system name'`

and probably to be safe `tags.sap_system = 'system name' and (name = 'system name' and type = 'SAP_SYSTEM' )`

I don't love the idea of having tags of the entity name but i can see the use case you are looking for. I'll let it go for now unless it becomes a problem in the future.

entityBot (approved):

PR #417, opened 2022-01-25
Add definition files for new entity type ext-sap_httpdest. This entit…

edyu002:
### Relevant information

Add definition files for new entity type ext-sap_httpdest. Http dest are defined in SAP system to track remote http end points.

### Checklist

* [x ] The value of the attribute marked as `identifier` will be unique and valid. 
 explanation above.

### DIFF
--- a/definitions/ext-sap_httpdest/dashboard.json
+++ b/definitions/ext-sap_httpdest/dashboard.json
@@ -1,0 +1,116 @@
+{
+  "name": "E-HTTPDEST",
+  "description": null,
+  "pages": [
+    {
+      "name": "RFC",
+      "description": null,
+      "widgets": [
+        {
+          "visualization": {
+            "id": "viz.line"
+          },
+          "layout": {
+            "column": 1,
+            "row": 1,
+            "height": 3,
+            "width": 6
+          },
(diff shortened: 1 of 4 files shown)


entityBot (approved):

danipl (approved):

naxhh:
I remember on the calls that you diferentiate rfc from http. but they look very similar on the data shape.

Would it make sense to have one entity type SAP_DESTINATION only and have a tag to differentiate http from the other one? does that makes sense?

edyu002:
Yes. On the golden metrics  and  entity type dashboard, the two types are
very similar. The difference of the two destinations are buried deeper in
more detailed data attributes. I'll discuss with the SAP team to see if it
is ok to merge those two types.

edyu002:
Follow up on this topic. I discussed this with our SAP basis people. Their
point is that in SAP, those 2 types of destinations are treated quite
differently, and datasource for them are different too. Even though the
golden metrics and dashboard are similar now, they will deviate when we add
entities into relationships context and more details are shown. So that
means we'd like to keep them as separate entity types.

entityBot (approved):

PR #420, opened 2022-01-25
Add definitions files for new entity type ext-maxdb.

edyu002:
### Relevant information

Add definition files for new entity type ext-maxdb. Maxdb is a type of database SAP uses.

### Checklist

* [x ] The value of the attribute marked as `identifier` will be unique and valid. 
 explanation above.

### DIFF
--- a/definitions/ext-maxdb/dashboard.json
+++ b/definitions/ext-maxdb/dashboard.json
@@ -1,5 +1,5 @@
 {
-  "name": "E-MAXDB",
+  "name": "EXT-MAXDB",
   "description": null,
   "pages": [
     {
--- a/definitions/ext-maxdb/dashboard.json
+++ b/definitions/ext-maxdb/dashboard.json
@@ -1,0 +1,119 @@
+{
+  "name": "E-MAXDB",
+  "description": null,
+  "pages": [
+    {
+      "name": "RFC",
+      "description": null,
+      "widgets": [
+        {
+          "visualization": {
+            "id": "viz.line"
+          },
+          "layout": {
+            "column": 1,
+            "row": 1,
+            "height": 3,
+            "width": 6
+          },
+          "title": "Trend - DB Data Space Usage %",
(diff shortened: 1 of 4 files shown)


entityBot (approved):

definitions/ext-maxdb/dashboard.json
```diff
@@ -0,0 +1,119 @@
+{
+  "name": "E-MAXDB",
```
naxhh, line 2:
```suggestion
  "name": "EXT-MAXDB",
```

definitions/ext-maxdb/definition.yml
```diff
@@ -0,0 +1,29 @@
+domain: EXT
+type: MAXDB
```
naxhh, line 2: should we prefix this with sap to keep consistent or there is a reason to not do that?

naxhh (approved):

definitions/ext-maxdb/definition.yml
edyu002: Maxdb is a database. During the discussion with the entity-platform team, it is indicated to us that it is better to keep the database entity type generic, such as reusing the existing  ext-db2 type if we can. So we are proposing the new DB entity types such as EXT_MAXDB, EXT-HANADB, EXT-ORACLE, and EXT-SQLSERVER.  Ext-MAXDB is the first one being generated, the rest will come later when we have customers using those databases.

definitions/ext-maxdb/dashboard.json
edyu002: Good catch. I'll try to make the dashboard name to be consistent with the entity type.

hredondo (approved):

definitions/ext-maxdb/definition.yml
naxhh: sounds great!

entityBot (approved):

PR #436, opened 2022-02-07
Add NPM Entities

Mesverrum:
### Relevant information

Created new entities for the following NPM objects:
ext-air-conditioner
ext-ipam
ext-web-gateway

Updated the following pre-existing NPM entities:
ext-kentik-default
ext-load_balancer
ext-nas
ext-wireless_controller

### Checklist

* [x] I've read the guidelines and understand the acceptance criteria.
* [x] The value of the attribute marked as `identifier` will be unique and valid. 
* [x] I've confirmed that my entity type wasn't already defined. If it is I'm providing an
 explanation above.

PR #437, opened 2022-02-09
fix/kentik entity dashboards

thezackm:
### Relevant information

Dashboard fixes applied to items from PR #436 

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-air_conditioner/liebert-dashboard.json
+++ b/definitions/ext-air_conditioner/liebert-dashboard.json
@@ -18,11 +18,17 @@
             },
             "title": "Summary",
             "rawConfiguration": {
-              "dataFormatters": [],
+              "dataFormatters": [
+                {
+                  "name": "Uptime (Days)",
+                  "precision": 2,
+                  "type": "decimal"
+                }
+                ],
               "nrqlQueries": [
                 {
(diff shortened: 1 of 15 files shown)


entityBot (approved):

danipl (approved):

naxhh (approved):
we probably want to double check that data formatter works since i think we have never used them (they should work afaik)

thezackm:
> we probably want to double check that data formatter works since i think we have never used them (they should work afaik)

good eye 😉 

I've used them with the dashboards API a lot so they're good there; and I can see them on a few existing entity dashboards like the [ext-deeper_connect](https://github.com/newrelic/entity-definitions/blob/02df6bd0ec33f0ba98e1047a5368b17525621bc3/definitions/ext-deeper_connect/dashboard.json#L206-L227) and [ext-host](https://github.com/newrelic/entity-definitions/blob/02df6bd0ec33f0ba98e1047a5368b17525621bc3/definitions/ext-host/flex-dashboard.json#L143-L169), along with some of the pre-existing entities in the network space.

naxhh:
@thezackm great! i had this idea that this was a new thing but happy to be corrected 🎉

PR #448, opened 2022-02-15
Storage Entity Types

khpeet:
### Relevant information

Adding entity definitions for the following storage types - all telemetry is emitted via custom Flex configurations that is unique to a single NR account.

- Dell Compellent
- DellEMC Isilon
- Pure
- Dell RecoverPoint (RPA)
- Synology
- DellEMC VMAX
- DellEMC VNX
- DellEMC VPLEX
- DellEMC XtremIO

### Checklist

 explanation above.


definitions/ext-dell_compellent/definition.yml
```diff
@@ -0,0 +1,39 @@
...
+type: DELL_COMPELLENT
+
+synthesis:
+  name: clusterCI
+  identifier: clusterCI
+  encodeIdentifierInGUID: true
+  conditions:
+  - attribute: eventType
+    prefix: compellent_
+  - attribute: clusterCI
```
naxhh, line 11: We already check that the attribute for the name or the indetifier exists in the data points so this condition is not needed

definitions/ext-dell_compellent/summary_metrics.yml
```diff
@@ -0,0 +1,7 @@
+diskCount:
```
naxhh, line 1: As we said in the previous PR can you use the golden metrics her too? that will make the migration much easier for everyone.

definitions/ext-dell_compellent/golden_metrics.yml
```diff
@@ -0,0 +1,24 @@
...
+failedDriveCount:
+  title: Failed Drive Occurences
+  unit: COUNT
+  query:
+    select: count(status_code) as 'Drive Failures'
+    from: compellent_disks
+    where: status = 'Down'
+  displayAsValue: true
+
+usedCapacityPercent:
```
naxhh, line 18: this will be the name to query by, for example:

`select newrelic.goldenmentric.ext.dell_compellent.usedCapacityPercent FROM Metric` not sure if we want to go with that or without the percent part

naxhh (commented):
Comments apply for all the new types.

Looking at this PR's I still think this should be one entityType (most probably ext-nas) with golden metrics like cpu, ram & usedCapacity

Edit: we are moving forward with this but please first address the feedback

naxhh:
we would need the sla signed with the 2nd user too

definitions/ext-dell_compellent/golden_metrics.yml
khpeet: thats fine

definitions/ext-dell_compellent/summary_metrics.yml
khpeet: done

entityBot (approved):

definitions/ext-dell_compellent/definition.yml
khpeet: removed second condition from all definitions

khpeet:
> 

I've signed with both - though it's unnecessary since both GH users are me. I've merged PRs before without having to sign both.

naxhh:
```
../definitions/ext-dell_compellent/summary_metrics.yml invalid
[
  {
    "instancePath": "/failedDriveCount/query",
    "schemaPath": "#/definitions/query/required",
    "keyword": "required",
    "params": {
      "missingProperty": "eventId"
    },
    "message": "must have required property 'eventId'"
  }
]
```

entityBot (approved):

khpeet:
image

naxhh (approved):

apoloa (approved):

entityBot (approved):

PR #465, opened 2022-02-25
Add some Azure entity synthesis rules

morecar:
### Relevant information

Adding some synthesis rules for 3 Azure resourceTypes

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-azureapimanagementservice/definition.yml
+++ b/definitions/infra-azureapimanagementservice/definition.yml
@@ -8,3 +8,11 @@ goldenTags:
 configuration:
   entityExpirationTime: DAILY
   alertable: true
+synthesis:
+  rules:
+  - identifier: azure.resourceId
+    name: displayName
+    encodeIdentifierInGUID: true
+    conditions:
+    - attribute: azure.resourceType
+      value: Microsoft.ApiManagement/service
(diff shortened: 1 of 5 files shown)


entityBot (approved):

definitions/infra-azureapimanagementservice/definition.yml
```diff
@@ -8,3 +8,11 @@ goldenTags:
 configuration:
   entityExpirationTime: DAILY
   alertable: true
+synthesis:
+  rules:
+  - identifier: azure.resourceId
```
naxhh, line 13: Just for my knowledge. what kind of values we get here? similar to aws arn's?

naxhh (approved):

definitions/infra-azureapimanagementservice/definition.yml
morecar: It-s unique like an ARN, but it's built more like the path in a file system, with slashes. The structure is something like:
```
/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.{serviceName}/{resourceType}/{resourceName}
```

form that, the substring `Microsoft.{serviceName}/{resourceType}` would be your `resourceType`.

entityBot (approved):

apoloa (approved):

entityBot (approved):

PR #484, opened 2022-03-11
fix: add entity.guid to widget

Mesverrum:
### Relevant information

Added required for the entity.guid filtering that gets applied when viewing a specific entity

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-kentik_default/dashboard.json
+++ b/definitions/ext-kentik_default/dashboard.json
@@ -314,7 +314,7 @@
 			  "nrqlQueries": [
 				{
 				  "accountId": 0,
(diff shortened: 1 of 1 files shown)


entityBot (approved):

naxhh (approved):
I just noticed this is using a query inside a query  :\

Our very straightforward filtering doesn't take that into account and hence you need to provide a facet (since the inner query will return a bunch of entities) and later filter the one you actually want.

is that what is going on here?

Mesverrum:
> 

Yes exactly, the purpose is to draw attention to volumes that have high utilization

entityBot (approved):

PR #492, opened 2022-03-17
fix: update netapp instrumentation.name and cleanup SM and GM

thezackm:
### Relevant information

* Corrected a typo on the `instrumenation.name` for NetApp clusters
* Removed deprecated tab from SM
* Removed memory util from SM and GM as it's not a standard metric across NAS devices
* Updated CPU util in SM and GM to add the NetApp clusters

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-nas/golden_metrics.yml
+++ b/definitions/ext-nas/golden_metrics.yml
@@ -1,16 +1,12 @@
-# Tested on Netgear ReadyNAS
 cpuUtilization:
-  title: CPU Utilization (%)
+  title: CPU
   unit: PERCENTAGE
-  query:
-    select: average(kentik.snmp.CPU)
-    from: Metric
-    where: "provider = 'kentik-nas'"
-
-memoryUtilization:
-  title: Memory Utilization (%)
-  unit: PERCENTAGE
-  query:
-    select: average(kentik.snmp.MemoryUtilization)
-    from: Metric
-    where: "provider = 'kentik-nas'"
+  queries:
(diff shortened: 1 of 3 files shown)


entityBot (approved):

definitions/ext-nas/golden_metrics.yml
```diff
@@ -1,16 +1,12 @@
...
-  unit: PERCENTAGE
-  query:
-    select: average(kentik.snmp.MemoryUtilization)
-    from: Metric
-    where: "provider = 'kentik-nas'"
+  queries:
+    kentik/readynas:
+      select: 100 - latest(kentik.snmp.ssCpuIdle)
+      from: Metric
+      where: "provider = 'kentik-nas' AND SysObjectID = '.1.3.6.1.4.1.4526.100.16'"
```
otaviocarvalho, line 8: They always have the same SysObjectID?

otaviocarvalho (approved):

hredondo (approved):

definitions/ext-nas/golden_metrics.yml
hredondo: We'll merge once confirmed!

thezackm: The Netgear ReadyNas do, that's been in the definition since day 1. We need to refactor this but we don't have any customers sending us this data to test against right now unfortunately so it's best to leave it as is IMO.

PR #496, opened 2022-03-23
feat: add additional tags and event type to EXT-HOST

Kav91:
### Relevant information

Adds additional event type and tags to the EXT-HOST synthesis.
The data is modelled identically to the `FlexSystemSample` but generated via ansible.

PR is required for the Cloud Accelerator project.

### Checklist

 explanation above.


entityBot (approved):

entityBot (approved):

definitions/ext-host/definition.yml
```diff
@@ -17,6 +17,41 @@ synthesis:
...
+      conditions:
+      - attribute: eventType
+        value: AnsibleServiceSample
+    - identifier: displayName
+      name: displayName
+      encodeIdentifierInGUID: true
+      conditions:
+      - attribute: eventType
+        value: AnsibleStorageSample 
+    - identifier: displayName
```
naxhh, line 48: without a custom dashboard for this they will use the first one defined for datadog for the summary view.
And I don't think that will match on telemetry right?

naxhh (commented):
EXT-HOST has been largely used for entities that are temporary for transition. Is that also the intention here?

If we want to treat these ones as "first class" hosts they should be under infra-host where all hosts are.

definitions/ext-host/definition.yml
Kav91: @naxhh correct, for context in the cloud accelerator project we are using the data generated from Flex and Ansible to create entities which will then eventually be migrated to the cloud and have our core agents in the landing zone. We are using workloads to track the entities between on-prem and cloud.

Re. custom dashboard we are fine with not having any dashboard associated to the other telemetry, we just want the ensure the same guid across the event types.

entityBot (approved):

entityBot (approved):

entityBot (approved):

naxhh (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

danipl (approved):

danipl (approved):

definitions/ext-host/definition.yml
danipl: I don't have context enough to understand the deep explanation, @naxhh or @Kav91, could any of you confirm the approach reached by the PR is the desired?

Kav91: @danipl yes it should produce our desired outcome. Also as the usage is scoped to the cloud accelerator project and our build partner, it will not have any large customer impact at this stage.

PR #507, opened 2022-03-29
Improves synthesis doc.

danipl:
### Relevant information

Improves the synthesis documentation, documenting different ways of copying tags on an entity.

### Checklist

 explanation above.

### DIFF
--- a/docs/synthesis.md
+++ b/docs/synthesis.md
@@ -141,7 +141,7 @@ synthesis:
 
 If the attribute `aws.az` is present in the data point that matched any rule, its value will be copied to an entity's tag named `aws.az`.
 
-And also there are specific tags for every rule. These attributes will only be copied after a datapoint matches a rule, and
+Also, there specific tags for each rule can be defined. These attributes will only be copied after a datapoint matches a rule.
 they take priority over the global tags, meaning **if the same tag is defined on both structures**, the specific tag  
 overrides the global tag.
 
--- a/docs/synthesis.md
+++ b/docs/synthesis.md
@@ -142,7 +142,7 @@ synthesis:
 If the attribute `aws.az` is present in the data point that matched any rule, its value will be copied to an entity's tag named `aws.az`.
 
 Also, there specific tags for each rule can be defined. These attributes will only be copied after a datapoint matches a rule.
-they take priority over the global tags, meaning **if the same tag is defined on both structures**, the specific tag  
+these rule specific tags take priority over the global tags, meaning **if the same tag is defined on both structures**, the rule specific tag  
 overrides the global tag.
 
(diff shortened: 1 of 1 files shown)


entityBot (approved):

docs/synthesis.md
```diff
@@ -128,18 +128,34 @@ synthesis:
 
 ### Tags
 
-You can also define a set of attributes that can be copied into entity tags. These attributes will only be copied after a datapoint matches a rule.
+You can also define a set of attributes that can be copied into entity tags.
 
+These are global tags, and they are applied for the rule or every rule from a set of rules.
```
rodolfocal, line 133:
```suggestion
These are global tags, and they are applied for all the rules from the set of rules.
```

docs/synthesis.md
```diff
@@ -128,18 +128,34 @@ synthesis:
...
+Notice these tag definitions live under ```synthesis```.
+```yaml
+synthesis:
+  tags:
+    aws.az:
+```
+
+If the attribute `aws.az` is present in the data point that matched any rule, its value will be copied to an entity's tag named `aws.az`.
+
+And also there are specific tags for every rule. These attributes will only be copied after a datapoint matches a rule, and
```
rodolfocal, line 144:
```suggestion
Also, there specific tags for each rule can be defined. These attributes will only be copied after a datapoint matches a rule.
```

docs/synthesis.md
```diff
@@ -128,18 +128,34 @@ synthesis:
...
+they take priority over the global tags, meaning **if the same tag is defined on both structures**, the specific tag
```
rodolfocal, line 145:
```suggestion
these rule specific tags take priority over the global tags, meaning **if the same tag is defined on both structures**, the rule specific tag  
```

rodolfocal (approved):
Thanks @danipl , it is much clear now. I suggested some changes in the text.

naxhh (approved):

apoloa (approved):

docs/synthesis.md
danipl: Just to notice, maybe I need to rephrase it, but I would like to remark two approaches.
1. There's a unique rule, where the rule itself lives inherited and directly under "synthesis:".
2. There're multiple rules, where they live under "rules:".

entityBot (approved):

entityBot (approved):

entityBot (approved):

docs/synthesis.md
danipl: Improved, I think it's much clearer like this.

entityBot (approved):

entityBot (approved):

PR #503, opened 2022-03-25
Add External entities

lucidobservor:
### Relevant information

Infinite Tracing produces DistributedTraceProcessSummary events. These DistributedTraceProcessSummary events include unmonitored externals. In the case of these unmonitored externals, we basically “synthesize” an entity that represents that url.

So we’ve got all of these DistributedTraceProcessSummary events out there and we’ve synthesized these entity guids and attached them to those events. However, since these entities are not monitored with New Relic, they are never indexed by the entity platform and so you can’t search on them or filter by them, etc.

We _believe_ adding this synthesis definition will allow Entity Platform to start indexing the data contained in those DistributedTraceProcessSummary events.

Some Slack discussion in #external-entities

### Checklist

* [x] I've read the guidelines and understand the acceptance criteria.
* [x] The value of the attribute marked as `identifier` will be unique and valid. 
* [x] I've confirmed that my entity type wasn't already defined. If it is I'm providing an
 explanation above.

PR #517, opened 2022-03-31
Add external service

lucidobservor:
### Relevant information

This is an updated version of this PR: https://github.com/newrelic/entity-definitions/pull/503

In the previous PR, I made a lot of commits where the `user.name` and `user.email` were not associated with a public github account, meaning I could not accept the CLA. This PR rectifies this.

We (the DT team) believe that this PR, in conjunction with PR 107 on internal service, should suffice to start materializing the unmonitored external entities sent in DistributedTraceProcessSummary records. I do not believe the order in which those tow PRs are merged will matter.

The previous PR previously contained test files. It is my belief that the current test setup is unable to handle this configuration for reasons unrelated to its validity. Perhaps the EXTERNAL domain needs to be added to the test setup?

### Checklist

 explanation above.


entityBot (approved):

definitions/external-service/definition.yml
```diff
@@ -0,0 +1,16 @@
+domain: EXTERNAL
+type: SERVICE
+
+synthesis:
+  rules:
+    # DistributedTraceProcessSummary events
+    - identifier: entity.guid
```
hredondo, line 7: This is something that I don't think is what you would expect. This represents the domain id. See [this doc](https://pages.datanerd.us/site-engineering/nr-platform-docs/entity-platform/entity-indexing/entity-indexing.html#entity-global-unique-identifier) for more information. It's not the guid that the synthesized entity would have, but just one of the 4 sections inside an entity guid. Same for the name.

You need to find which is the name and id that are coming along in the telemetry, and use those attributes as the source of the potential entity guid and entity name.

hredondo (changes requested):

definitions/external-service/definition.yml
jvanpeltnr: entity.guid on DistributedTraceProcessSummary is the complete, encoded entity guid.
The domain ID upon which that entity guid is based depends on the entity type. When the entity is an EXTERNAL SERVICE, the entity.name is the host name, which is used for the domain ID in the entity guid.

jvanpeltnr:
A given DistributedTraceProcessSummary event could be for one of many types of entities, including APM APPLICATION, EXT SERVICE, etc. EXTERNAL SERVICE is just one of those types. 

We only want to create an EXTERNAL SERVICE when a DistributedTraceProcessSummary event is specifically for that type of entity. (And we don't _need_ to create an entity for any other types at this point since any other type should already be synthesized via Span or other events)

a) How do we ensure the results described above?
b) Generally speaking, when entity.type is on an event, is it assumed to be in the EXT domain? Is that attribute used in this case, or only if entity.guid is not already defined on the event?

hredondo:
> A given DistributedTraceProcessSummary event could be for one of many types of entities, including APM APPLICATION, EXT SERVICE, etc. EXTERNAL SERVICE is just one of those types.
> 
> We only want to create an EXTERNAL SERVICE when a DistributedTraceProcessSummary event is specifically for that type of entity. (And we don't _need_ to create an entity for any other types at this point since any other type should already be synthesized via Span or other events)
> 
> a) How do we ensure the results described above? b) Generally speaking, when entity.type is on an event, is it assumed to be in the EXT domain? Is that attribute used in this case, or only if entity.guid is not already defined on the event?

Hi @jvanpeltnr, how the synthesis process works is, given the event, defining a rule that would match it. So, an entity of the definition where the rule is defined will be created. 

For your use case, I assume the entity.guid that is already present in `DistributedTraceProcessSummary` has `EXTERNAL-SERVICE` domain and type. This is the way we know we need to look up for a matching rule in the EXTERNAL-SERVICE definition. If that's not the case, how the synthesis works today does not suit your needs and we'd need to find a solution. If the entity.guid has many forms and you want eventually to always create a EXTERNAL-SERVICE entity, we'll need to find another way of doing it.

lucidobservor:
> I assume the entity.guid that is already present in `DistributedTraceProcessSummary` has `EXTERNAL-SERVICE` domain and type. 

Mostly correct. `DistributedTraceProcessSummary` events can have many domains and types, but EXTERNAL-SERVICE is the only one we want to materialize here.

> This is the way we know we need to look up for a matching rule in the EXTERNAL-SERVICE definition. 

So if I understand this right, 
- `synthesis-service` will begin receiving `DistributedTraceProcessSummary` events
- `synthesis-service` will look at the already-present and encoded `entity.guid` on those events to determine the Domain and Type
- For some subset of those events, domain and type will be EXTERNAL and SERVICE. This will cause synthesis service to look at the rule being defined in this PR, and materialize an entity with the defined `entity.guid`

That seems fine as far as it goes. For `DistributedTraceProcessSummary` whose domain and type are something _else_ I'm a bit fuzzier on what the behavior will be. In theory, those should already be synthesized by other sources (see Jason's comment). But do we need to worry about duplicates?

naxhh:
Synthesis engine goal is to identify an entity out of the telemetry being fed.
From your description it seems you don't need this part since you are already identifying it by yourself.

So what you are trying to do is use the synthesis-service as a materialization mechanism only.
This is not the expected usage of the service and hence our reluctancy as using it as such.

You could archive exactly the same sending the message directly into the internal topic.
I'm not sure I fully understand why you want to use synthesis in this case.

From other conversations I misunderstood that you where going to feed it unidentified telemetry and let it decide what to create (the intended usage).

I don't think we should use the rules for materialization only since makes the whole system more complex and we deviate from the standard.

Also even if we where to use the synthesis service the rule is still incorrect, since the `identifier` can't be the full guid but rather the piece that ends up in the domainId

lucidobservor:
Conversation moved to Slack: https://newrelic.slack.com/archives/C02H63LGN7R/p1649254386243509

lucidobservor:
Closing for now, since bypassing synthesis seems like it might be the better approach

entityBot (approved):

entityBot (approved):

naxhh (approved):
As we agreed even when merged the definition we will not create them yet since we need to encode the identifier field.

Thanks a lot for the patience on this one!

entityBot (approved):
beep boop bop.

I have found a debug rule, this is the output for this rule:

 ```
Debug Entity: 
SynthesizedEntity{domain='EXT', entityType='ACCESS_POINT', domainTypeType='null', accountId=1, domainId='3997086220620660013', name='192.168.1.253', singleValueTags='{}', multiValueTags='{instrumentation.name=remote-write, instrumentation.provider=prometheus}'}
Using the rule:
SynthesisRule{index='106', domain='EXT', entityType='ACCESS_POINT', domainType='null', domainIdSpec='SingleAttributeDomainId{attrName='serial'}', nameAttribute='ip', conditions='[com.newrelic.entitysynthesisengine.api.models.RuleCondition@fe718845, com.newrelic.entitysynthesisengine.api.models.RuleCondition@83c8f05d]', mustHashDomainId=true} 
```

otaviocarvalho (approved):

PR #527, opened 2022-04-06
feat: Adding a practice entity

andre-nr:
### Relevant information

Just adding a practice entity for onboarding purposes.

### Checklist

 explanation above.

### DIFF
--- a/definitions/test-andre/definition.yml
+++ b/definitions/test-andre/definition.yml
@@ -1,0 +1,17 @@
+domain: TEST
+type: ANDRE
+
+synthesis:
+  rules:
+    - identifier: someIdField
+      encodeIdentifierInGUID: true
+      name: someNameField
+      conditions:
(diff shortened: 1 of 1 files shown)


entityBot (approved):

definitions/test-andre/definition.yml
```diff
@@ -0,0 +1,17 @@
+domain: TEST
+type: ANDRE
+
+synthesis:
```
naxhh, line 4: since this is for testing and you don't need the synthesis capabilities you can remove all this section.

If you are planning to actuall test this I said nothing and you can keep it.
Maybe just in case change the metric name to something even more unique like: andre_testing_1234g

naxhh (approved):

otaviocarvalho (approved):

entityBot (approved):

PR #547, opened 2022-04-18
fix/network gm queries

thezackm:
### Relevant information

updating several network entities to support "default"/"catch-all" rule for the GM Harvester service. This change removes some of the `instrumentation.provider/instrumentation.name` patterns for GM and SM definitions and adds the more generic `instrumentation.provider` pattern to capture all telemetry for an entity type coming from Kentik's ktranslate container.

### Checklist

 explanation above.


entityBot (approved):

otaviocarvalho (approved):
Thanks for moving this forward, @thezackm!

naxhh (approved):

PR #628, opened 2022-06-08
NR-17236 fix SM GM infra msqlite vshpereresourcepool

albertoolivan:
### Relevant information

Same as other PR with SummaryMetric a subset of GoldenMetric: infra msqlite and vsphereresourcepool

### Checklist

 explanation above.


danipl (approved):

definitions/infra-vsphereresourcepool/golden_metrics.yml
```diff
@@ -54,3 +54,17 @@ memoryFree:
...
+  query:
+    eventId: entityGuid
+    select: latest(`vmCount`)
+    from: VSphereResourcePoolSample
+  unit: COUNT
+  title: Virtual machines count
+memSwapped:
+  query:
+    eventId: entityGuid
+    select: (average(`mem.swapped`)) * 1024 * 1204
```
gvassallo, line 67: given that the unit is BYTES, is this `* 1024 * 1024` actually correct?

albertoolivan: It makes sense, this query was copied from https://github.com/newrelic/entity-definitions/blob/c153c727ec0aa87543018927885636082ae7f348/definitions/infra-vsphereresourcepool/summary_metrics.yml#L11 

After some investigation, this query  shows https://onenr.io/0BQrK3JoNwZ
select (average(`mem.free`)) * 1024, (average(`mem.free`)) * 1024 * 1024 FROM VSphereResourcePoolSample
4.83 G
4.94 T -> it seems mistaken

mangulonr:
From Core Integrations we approve the PR meaning the golden metrics you want to add.
But there is missing the equivalent query for DMs in each one of them.

Each Golden metric has two queries (one for DM one for samples)

Thanks @albertoolivan

apoloa (approved):

PR #635, opened 2022-06-14
point to golden metrics for awsalb

vetusbs:
This will allow the metrics to be fetched from the same backend as GM.


albertoolivan:
ok jenkins

PR #668, opened 2022-07-05
added definitions for Mongo metrics

kfalconer:
### Relevant information

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-mongo/definition.yml
+++ b/definitions/infra-mongo/definition.yml
@@ -1,0 +1,5 @@
+domain: INFRA
+type: MONGO
+configuration:
+  entityExpirationTime: DAILY
+  alertable: true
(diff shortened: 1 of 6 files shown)


kfalconer:
> hi @kfalconer ! can you give us a bit more context on this?
> 
> This PR will create the types but how is the data going to be created & managed

I am not sure I understand the question. Do you mean source of the metrics? This is the coming from the [nri-mongo](https://github.com/newrelic/nri-mongodb) integration.

I followed the same pattern as the other Infrastructure based integrations by defining the types. But it wasn't clear on the examples how the data is actually synthesized since there was no examples of that.

naxhh:
@kfalconer I see what you mean.

Let me run down this on the teams providing this integrations.

The thing is that Infra entities go to a different pipeline and they don't use synthesis (yet) so manual work needs to be done in order to create such entities.

In the future these will go through synthesis and the process of creating entities will be simpler

naxhh:
@kfalconer bringing more info from the team!

The work to bring these entities into the platform is already prioritized on the team backlog.

So i'll close this PR and let them do the required work and create the types.

Thanks a lot for your contribution!

kfalconer:
Thanks for the update

PR #678, opened 2022-07-12
remove rabbitmq cluster entity

gsanchezgavier:
### Relevant information
Removing the Rabbitmq cluster entity so it's no longer synthetized and present in the platform.

We have deprecated this entity and[ EOL anouncement](https://discuss.newrelic.com/t/q1-bulk-eol-announcement-fy23/181744)

The entity didn't have any metrics by it's own but was just using metrics from other entities. The charts were broken and fixing them for Dimensional Metrics was not possible.

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-rabbitmqcluster/definition.yml
+++ b/definitions/infra-rabbitmqcluster/definition.yml
@@ -1,7 +1,0 @@
-domain: INFRA
-type: RABBITMQCLUSTER
-goldenTags:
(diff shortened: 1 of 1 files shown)


entityBot (approved):

naxhh (approved):
We should check how many entities we have in the storage before merging this one.

LGTM regardless.

PR #680, opened 2022-07-13
fix some queries of sm gm

albertoolivan:
### Relevant information

Checking discrepancies between Summary metrics results in some entities with GoldenMetrics. Found some query discrepancies and minor errors.

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-fibre_channel_switch/golden_metrics.yml
+++ b/definitions/ext-fibre_channel_switch/golden_metrics.yml
@@ -5,6 +5,7 @@ cpuUtilization:
     select: average(kentik.snmp.CPU)
     from: Metric
     where: "provider = 'kentik-fibre-channel'"
+    eventId: entity.guid
 
 memoryUtilization:
   title: Memory utilization (%)
@@ -13,6 +14,7 @@ memoryUtilization:
     select: average(kentik.snmp.MemoryUtilization)
     from: Metric
     where: "provider = 'kentik-fibre-channel'"
+    eventId: entity.guid
 
 portStatus:
   title: Offline ports
@@ -21,3 +23,4 @@ portStatus:
     select: uniqueCount(port_index)
     from: Metric
     where: "provider = 'kentik-fibre-channel' AND fcFxPortPhysOperStatus !='online'"
+    eventId: entity.guid
--- a/definitions/infra-awskinesisdeliverystream/golden_metrics.yml
+++ b/definitions/infra-awskinesisdeliverystream/golden_metrics.yml
@@ -18,7 +18,7 @@ getThroughput:
   unit: BYTES_PER_SECOND
   queries:
     aws:
-      select: sum(aws.firehose.DeliveryToElasticsearch.Bytes) + sum(aws.firehose.DeliveryToS3.Bytes) + sum(aws.firehose.DeliveryToRedshift.Bytes)
+      select: rate(sum(aws.firehose.DeliveryToElasticsearch.Bytes) + sum(aws.firehose.DeliveryToS3.Bytes) + sum(aws.firehose.DeliveryToRedshift.Bytes), 1 second)
       from: Metric
       eventId: entity.guid
       eventName: entity.name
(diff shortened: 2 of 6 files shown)


entityBot (approved):

definitions/ext-fibre_channel_switch/golden_metrics.yml
```diff
@@ -13,6 +14,7 @@ memoryUtilization:
     select: average(kentik.snmp.MemoryUtilization)
     from: Metric
     where: "provider = 'kentik-fibre-channel'"
+    eventId: entity.guid
```
gvassallo, line 17: that is not really required, as it's the default value for the eventId

definitions/infra-awskinesisdeliverystream/golden_metrics.yml
```diff
@@ -18,7 +18,7 @@ getThroughput:
   unit: BYTES_PER_SECOND
   queries:
     aws:
-      select: sum(aws.firehose.DeliveryToElasticsearch.Bytes) + sum(aws.firehose.DeliveryToS3.Bytes) + sum(aws.firehose.DeliveryToRedshift.Bytes)
+      select: rate(sum(aws.firehose.DeliveryToElasticsearch.Bytes) + sum(aws.firehose.DeliveryToS3.Bytes) + sum(aws.firehose.DeliveryToRedshift.Bytes), 1 second)
```
gvassallo, line 21: What is the reason for this change? The queries from all the instrumentation providers need to use the same value function and as well have to use the same rate. We can't just add the rate function to one of them

definitions/infra-nginxserver/golden_metrics.yml
```diff
@@ -27,16 +27,30 @@ activeConnections:
...
-  query:
-    eventId: entityGuid
-    select: average(`net.connectionsAcceptedPerSecond`)
-    from: NginxSample
+  queries:
+    newRelic:
+      select: average(nginx.server.net.connectionsAcceptedPerSecond)
+      from: Metric
+      eventId: entity.guid
+      eventName: entity.name
```
gvassallo, line 35: Given that you're using the default values here, I don't think it's useful to set these values explicitly

definitions/infra-awskinesisdeliverystream/golden_metrics.yml
albertoolivan: @josemore is informed of that, the query is different between GM and SM and generate discrepancies

definitions/ext-fibre_channel_switch/golden_metrics.yml
albertoolivan: It is for set same query values as the rest.

definitions/infra-awskinesisdeliverystream/golden_metrics.yml
gvassallo: My point is that if you use a rate for one instrumentation provider, you'll have to use a rate for the other instrumentation providers as well. Otherwise, we might get different results based on whether the entity uses one or the other instrumentation provider.

definitions/infra-nginxserver/golden_metrics.yml
gvassallo: What I'm saying is that, as it's stated [here](https://github.com/newrelic/entity-definitions/blob/main/docs/golden_metrics.md#query-options) there are default values for the query/queries configuration objects. 
The default value for `from` is `Metric`, for `eventId` is `entity.guid` and for `eventName` is `entity.name`. 
For this reason, you don't need to override them on lines 33-35.

definitions/infra-awskinesisdeliverystream/golden_metrics.yml
albertoolivan: The idea is to have GoldenMetric with rate, same as their summary-metrics https://github.com/newrelic/entity-definitions/blob/main/definitions/infra-awskinesisdeliverystream/summary_metrics.yml#L18

entityBot (approved):

gvassallo (approved):

PR #682, opened 2022-07-13
fix sm gm awsroute53

albertoolivan:
### Relevant information

fix sm gm awsroute53

Added Sample query because in some cases GoldenSignals need it.

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-azurerediscacheshard/summary_metrics.yml
+++ b/definitions/infra-azurerediscacheshard/summary_metrics.yml
@@ -15,7 +15,7 @@ operations:
   goldenMetric: operations
   query:
     eventId: entityGuid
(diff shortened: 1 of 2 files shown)


entityBot (approved):

definitions/infra-azurerediscacheshard/summary_metrics.yml
```diff
@@ -15,7 +15,7 @@ operations:
   goldenMetric: operations
   query:
     eventId: entityGuid
-    select: average(`operationsPerSecond.Maximum`)
+    select: max(`operationsPerSecond.Maximum`)
```
albertoolivan, line 18: Confirmed by @josemore

gvassallo (approved):

PR #693, opened 2022-07-18
feat: Add support for GCP VPCs

RyanKadri:
### Relevant information

This PR adds support for synthesizing VPC entities from GCP VPC Flow Logs. I tried to follow the same pattern that AWS Flow Logs use but things are a little more awkward in GCP for a couple reasons:

1. VPC IDs are not annotated onto flow logs like in GCP. As a result, I had to make the entity identifier the VPC name and the project ID. GCP Project IDs [must be globally unique](https://cloud.google.com/resource-manager/docs/creating-managing-projects#:~:text=A%20globally%20unique%20identifier%20for%20your%20project) and then VPC names [must be unique in a project](https://cloud.google.com/vpc/docs/create-modify-vpc-networks#:~:text=Each%20new%20network%20that%20you%20create%20must%20have%20a%20unique%20name%20within%20the%20same%20project). Also, VPC names cannot be changed after they are created. I don't love this approach but I think it's the best way to get things working?
2. GCP VPC Flow Logs report the source and destination VPC when applicable so I had to use a semi-awkward workaround to pick the "right" VPC to synthesize. I make sure to select the "reporting" VPC's by using a condition on the `jsonPayload.reporter` and adding two rules for synthesis. LMK if you think there's a better approach.

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-vpc_network/definition.yml
+++ b/definitions/ext-vpc_network/definition.yml
@@ -16,14 +16,41 @@ synthesis:
       conditions:
         - attribute: provider
           value: firehose-vpc
+    # gcp pubsub -> dataflow (DEST)
+    - compositeIdentifier:
+        separator: "/"
+        attributes:
+          - jsonPayload.dest_vpc.project_id
+          - jsonPayload.dest_vpc.vpc_name
+      name: jsonPayload.dest_vpc.vpc_name
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: jsonPayload.reporter
+          value: "DEST"
+    # gcp pubsub -> dataflow (SRC)
+    - compositeIdentifier:
+        separator: "/"
+        attributes:
+          - jsonPayload.src_vpc.project_id
+          - jsonPayload.src_vpc.vpc_name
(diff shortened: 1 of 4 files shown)


entityBot (approved):
beep boop bop.

I have found a debug rule, this is the output for this rule:

 ```
Debug Entity: 
SynthesizedEntity{domain='EXT', entityType='ACCESS_POINT', domainTypeType='null', accountId=1, domainId='3997086220620660013', name='192.168.1.253', singleValueTags='{}', multiValueTags='{instrumentation.name=remote-write, instrumentation.provider=prometheus}'}
Using the rule:
SynthesisRule{index='131', domain='EXT', entityType='ACCESS_POINT', domainType='null', domainIdSpec='SingleAttributeDomainId{attrName='serial'}', nameAttribute='ip', conditions='[com.newrelic.entitysynthesisengine.api.models.RuleCondition@fe718845, com.newrelic.entitysynthesisengine.api.models.RuleCondition@83c8f05d]', mustHashDomainId=true} 
```

entityBot (approved):

otaviocarvalho (approved):

PR #739, opened 2022-08-16
feature/add-aiops-issue

zvikah17:
feature/add-aiops-issue

### DIFF
--- a/definitions/aiops-issue/definition.yml
+++ b/definitions/aiops-issue/definition.yml
@@ -2,4 +2,4 @@ domain: AIOPS
 type: ISSUE
 configuration:
   entityExpirationTime: MANUAL
-  alertable: true
+  alertable: false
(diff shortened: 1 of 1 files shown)


entityBot (approved):

zvikah17:
see https://newrelic.slack.com/archives/CH5MLECNN/p1659435149051259

definitions/aiops-issue/definition.yml
```diff
@@ -0,0 +1,5 @@
+domain: AIOPS
+type: ISSUE
+configuration:
+  entityExpirationTime: MANUAL
+  alertable: true
```
andre-nr, line 5: I just wondered if issues are alertable or is this used to display them in the explorer ?

andre-nr (approved):

definitions/aiops-issue/definition.yml
zvikah17: As I understand it's required so issues could be searched in the explorer

apoloa: If this will not have an alert status should be set as `alertable: false`.

entityBot (approved):

apoloa (approved):

PR #743, opened 2022-08-23
chore: add 2 new telemetry sources to ext-ups

thezackm:
### Relevant information

adding 2 new data sources to the existing `EXT-UPS` entity definition

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-ups/golden_metrics.yml
+++ b/definitions/ext-ups/golden_metrics.yml
@@ -1,23 +1,59 @@
 outputLoadCapacity:
   title: Output load capacity %
   unit: PERCENTAGE
-  query:
-    select: average(kentik.snmp.upsAdvOutputLoad)
-    from: Metric
-    where: "provider = 'kentik-ups'"
+  queries:
+    # APC UPS profiles (default)
+    kentik/apc_ups:
+      select: max(kentik.snmp.upsAdvOutputLoad)
+      from: Metric
+      where: "provider = 'kentik-ups'"
+    # Tripplite UPS profiles
+    kentik/tripplite-ups:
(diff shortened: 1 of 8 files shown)


entityBot (changes requested):
beep boop bop.

 ```
* Error VALIDATING domainType: 'EXT-EMAIL_GATEWAY' on Golden metrics for key 'SELECT max(kentik.snmp.oldestMessageAge) FROM Metric WHERE entity.guid IN ('guid') AND provider = 'kentik-ironport-email-appliance' FACET entity.name TIMESERIES' - Exception: The select clause uses a different value function than the rest.
```

**Review by entityBot — APPROVED** [review]

File: definitions/ext-ups/golden_metrics.yml [hunk 1]
```diff
@@ -1,23 +1,59 @@
...
+    # APC UPS profiles (default)
+    kentik/apc_ups:
+      select: max(kentik.snmp.upsAdvOutputLoad)
+      from: Metric
+      where: "provider = 'kentik-ups'"
+    # Tripplite UPS profiles
+    kentik/tripplite-ups:
+      select: max(kentik.snmp.upsOutputPercentLoad)
+      from: Metric
+      where: "provider = 'kentik-ups'"
```
naxhh, line 14: is this correct? shouldn't be tripplite-ups? same for the previous one

naxhh (approved):
small question to make sure is correct but other than that lgtm

definitions/ext-ups/golden_metrics.yml
thezackm: It's accurate. Tripplite and the generic UPS have the same metrics as of right now

PR #747, opened 2022-08-26
Updated documentation for golden metric references

carl0FF:
- Updated the docs to reflect the new `goldenMetric` reference capability
- Removed mentions to `hidden` and `derive` since we're using them internally only
- Leaving for the time being `query`, mentioning its deprecation in favor of `goldenMetric`

### DIFF
--- a/docs/summary_metrics.md
+++ b/docs/summary_metrics.md
@@ -2,7 +2,7 @@
 
 Summary metrics are data related to entities which describe how certain parameters behave, based on the available telemetry. 
 
-We recommend adding 3 metrics, and never more of 10. We also recommend defining them like the golden metrics.
+We recommend adding 3 metrics, and never more than 10. We also recommend defining them as a reference to a golden metric.
 
 ## Defining summary metrics
 
@@ -10,6 +10,7 @@ Summary metrics are defined in a map where each key is a camel-case name that de
 
 ```yaml
 memoryUsage:
+  goldenMetric: memoryUsage
   title: "A title explaining what the user is seeing"
   unit: PERCENTAGE
   query:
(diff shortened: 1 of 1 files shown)


entityBot (approved):

entityBot (approved):

entityBot (approved):

docs/summary_metrics.md
```diff
@@ -19,16 +20,15 @@ memoryUsage:
     eventId: entity.guid
 ```
  
-A summary metric can be either an **NRDB-query-based metric**, a **tag metric** (summary tag), or a **derived metric**, depending on the source of the information. Therefore, you need to configure exactly ONE out of `query`, `tag` or `derive`. If an amount different to one is detected, the metric definition will not be valid.
+A summary metric can be either an **NRDB-query-based metric**, a **tag metric** (summary tag), or a **golden metric reference**, depending on the source of the information. Therefore, you need to configure exactly ONE out of `query` or `tag`. If an amount different to one is detected, the metric definition will not be valid. `query` can coexist with `goldenMetric` for the time being, but the former is being deprecated in favor of the latter.
```
otaviocarvalho, line 23: Do we have an estimated ETA for that? Should we only point users to goldenMetric instead?

rodolfocal (approved):

docs/summary_metrics.md
```diff
@@ -64,39 +64,30 @@ The `query` map contains the information used to define a NRDB-query-based metri
 | where         | String                         | An *optional* NRQL filter, for example `foo = 'bar'`.                                                  |
 | eventId       | String                         | The event attribute you want to facet and filter for, for example `entity.guid`.                           |
 
+**Note**: `query` is deprecated: use `goldenMetric` instead.
```
albertoolivan, line 67: Please query is deprecated on in summary definitions only

gvassallo (approved):

docs/summary_metrics.md
carl0FF: When all summary metrics are migrated to use `goldenMetric`, `query` support will be removed. This should happen soon as it is currently in progress.

docs/summary_metrics.md
carl0FF: 🤔 I think it's pretty clear the way it is - what exactly do you have in mind?

PR #753, opened 2022-09-05
Remove eventId and from as required fields in SM query configuration

gvassallo:
### Relevant information

The two fields are not required as, when not present, default values will be assigned for them. 

### Checklist

 explanation above.

### DIFF
--- a/validator/schemas/summary-metrics-schema-v1.json
+++ b/validator/schemas/summary-metrics-schema-v1.json
@@ -138,9 +138,7 @@
         "eventId": "entity.guid"
       }],
       "required": [
-        "select",
-        "from",
-        "eventId"
+        "select"
       ],
       "properties": {
         "select": {


entityBot (approved):

validator/schemas/summary-metrics-schema-v1.json
```diff
@@ -138,9 +138,7 @@
         "eventId": "entity.guid"
       }],
       "required": [
-        "select",
-        "from",
```
otaviocarvalho, line 142: Nice, it will match what we have in GMs now: https://github.com/newrelic/entity-definitions/blob/828b9b663ac3759bb17bbd93f9496e4418ed1c95/docs/golden_metrics.md#query-options

otaviocarvalho (approved):

naxhh (approved):

PR #767, opened 2022-09-21
Add new generic elastic cache entities

faltimiras:
### Relevant information

Create 2 new entities that englobes the elasticache (redis | memcached) cluster and node. 
This is needed because in some scenarios (Metric Streams) there is no way to distinguish between redis or memcached, then we need a way to somehow assign an entity to those metrics

### Checklist

 explanation above.


entityBot (approved):

andre-nr:
and this discussion: https://github.com/newrelic/entity-definitions/pull/765#issuecomment-1253341796

andre-nr (approved):

PR #775, opened 2022-09-27
Create windows-containers

iychan:
Identify events as entities in a windows docker environment

### Relevant information

We have instrumented the infrastructure agent with docker OHI on some Windows services. Events are in New Relic, we need a way to identify them as entities so that we can pull them into a workload.

### Checklist

* [x ] I've confirmed that my entity type wasn't already defined. If it is I'm providing an
 explanation above.

### DIFF
--- a/definitions/windows-containers
+++ b/definitions/windows-containers
@@ -1,0 +1,16 @@
+domain: Infra
+type: Windows_Containers
+
+synthesis:
+  rules:
+  
+  - identifier: hostname
+    encodeIdentifierInGUID: true
+    name: hostname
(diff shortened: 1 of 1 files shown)


entityBot (approved):

davidgit:
Hi @iychan ,

We appreciate you taking the time to propose a new entity definition. After reviewing this PR, we cannot accept this definition because these new synthesis rules would be applied to all the users sending the same telemetry. If the situation changes, we will revisit this decision.

If you have any further questions or comments about this, please let us know by replying to this issue.

PR #795, opened 2022-09-30
fix(firewall): Update metrics to use MemoryUtilization

jbeveland27:
### Relevant information

Some of the existing queries for summary/golden metrics were broken as they were performing a max operation on values that didn't sit on the same data point. This caused summary/golden metric values to be empty in the NR UI.

To fix, we added support in [kentik SNMP profiles](https://github.com/kentik/snmp-profiles/pull/435) to send Memory Utilization metrics and simplify the approach for generating golden metrics. 

This PR updates the corresponding summary/golden metric defs for Memory Utilization. It also has a fix for the `SessionTotal` metric calculation for fortigate firewalls.

### Checklist

 explanation above.


jbeveland27:
FYI: @Mesverrum

entityBot (approved):

definitions/ext-firewall/summary_metrics.yml
```diff
@@ -83,7 +71,7 @@ sessionsTotal:
       eventId: entity.guid
     # Fortinet Fortigate firewalls
     kentik/fortinet-fortigate:
-      select: max(kentik.snmp.fgSysSesCount + kentik.snmp.fgSysSes6Count)
```
gvassallo, line 86: could this be rewritten as `max(a + b)` instead of `max(a) + max(b)`? I believe the current query might lead to some issues on our side

entityBot (approved):

entityBot (approved):

PR #802, opened 2022-10-05
add redis replication group and shard entities

mkrichevsky:
### Relevant information

add redis replication group and shard entities

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-awselasticacheredisshard/golden_metrics.yml
+++ b/definitions/infra-awselasticacheredisshard/golden_metrics.yml
@@ -1,0 +1,10 @@
+engineCpuUtilization:
+  title: Engine CPU utilization (%)
+  unit: PERCENTAGE
+  queries:
+    aws:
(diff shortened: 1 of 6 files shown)


definitions/infra-awselasticacheredisshard/golden_metrics.yml
```diff
@@ -0,0 +1,10 @@
+engineCpuUtilization:
+  title: Engine CPU utilization (%)
+  unit: PERCENTAGE
+  queries:
+    aws:
```
gvassallo, line 5: are the metrics only reported as dimensional metrics or could they be reported as events too?

mkrichevsky: These ones are reported only as dimensional metrics

gvassallo (approved):

entityBot (approved):

PR #815, opened 2022-10-19
Add GMs for TLS version info by app

rcrocker:
Add golden metrics to report on how many instances are reporting supported vs deprecated versions of TLS. The definition of supported and deprecated versions will change over time, but providing this as a golden metric allows customers to plot their progress. This golden metric will be particularly useful as New Relic moves to deprecate a TLS version (as we're intending to do soon with TLS v1.0 and v1.1).

For example, a query like:
```
FROM Metric SELECT uniqueCount(newrelic.goldenmetrics.apm.application.deprecatedTlsInstances) TIMESERIES
```
will plot, across all applications in an account, the number of reporting instances using a deprecated version of TLS to talk to New Relic. Replacing `newrelic.goldenmetrics.apm.application.deprecatedTlsInstances` with `newrelic.goldenmetrics.apm.application.deprecatedTlsInstances` shows the number of reporting instances using a supported TLS version.

### Relevant information

### Checklist

 explanation above.


entityBot (approved):

rcrocker:
After thinking about this some more, this was a premature change. Closing for now.

PR #827, opened 2022-11-07
Run Sanitize dashboards in repo

philkuz:
### Relevant information
I'm developing a number of entity definitions right now. One of the required steps according to the instructions is to sanitize dashboards. I've found that when I run the command, these dashboards are changed as well. Adding these in to follow the standards set by the instructions.

Whoever is responsible for the CI system / entityBot should also probably reject PRs that don't satisfy the validation or remove that requirement.
### Checklist

 explanation above.


entityBot (approved):

entityBot (approved):

naxhh (approved):
Thanks!

Inter-team: we should have the conversation about what to do with those scripts..
The main idea was to provide them to cleanup not supported features or client data.

But i'm not really sure on how useful those are right now.

entityBot (approved):

entityBot (approved):

otaviocarvalho (approved):

PR #859, opened 2022-11-23
otel: temporary condition to use NA as entity type if special attr preset

rubenruizdegauna:
### Relevant information

### Checklist

 explanation above.

### DIFF
--- a/definitions/infra-host/definition.yml
+++ b/definitions/infra-host/definition.yml
@@ -128,6 +128,112 @@ synthesis:
       conditions:
         - attribute: azure.resourceType
           value: Microsoft.Compute/virtualMachines
+    # <test:entity_type_na>
+    # This is a temporary condition to set the entityType of the Otel Host to NA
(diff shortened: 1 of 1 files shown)


naxhh:
@rubenruizdegauna can you provide us more context on what is the intent here?

I think i can imagine it but would be good to provide context on the PR.

rubenruizdegauna:
> @rubenruizdegauna can you provide us more context on what is the intent here?
> 
> I think i can imagine it but would be good to provide context on the PR.

Hi @naxhh , sorry not to have added the context but I left it as a draft as it couldn't be merged at that time, and then I forgot about it :( . I'll close this as for now I can manage with [this definition](https://github.com/newrelic/entity-definitions/blob/main/definitions/infra-host/definition.yml#L123) and I'll reopen it properly if needed.

The goal was basically to be able to synthesize the HOST entity but as `NA` so I can have the Infra Agent and an Otel Collector reporting over the same host entity.

PR #880, opened 2022-12-28
Do not round sli compliance

NilVentosa:
### Relevant information

Make SLI compliance a STRING to avoid rounding to two decimals

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-service_level/summary_metrics.yml
+++ b/definitions/ext-service_level/summary_metrics.yml
@@ -12,8 +12,8 @@ sloPeriod:
 
 sliCompliance:
   goldenMetric: sliCompliance
-  title: SLI compliance
-  unit: PERCENTAGE
+  title: SLI compliance (%)
+  unit: STRING
 
 remainingErrorBudget:
   goldenMetric: remainingErrorBudget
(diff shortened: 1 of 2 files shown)


entityBot (approved):

entityBot (approved):

definitions/ext-service_level/summary_metrics.yml
```diff
@@ -12,10 +12,10 @@ sloPeriod:
 
 sliCompliance:
   goldenMetric: sliCompliance
-  title: SLI compliance
-  unit: PERCENTAGE
+  title: SLI compliance (%)
+  unit: STRING
```
gvassallo, line 16: Metrics should not be strings. only tags can be strings. what is the reason for this change?

NilVentosa: we would like the rounding to not be done to 2 digits as we allow SLO to have up to 4 decimal digits. We thought that by using a string the rounding would not take effect

gvassallo: I see, that seems to be a presentational issue that should be dealt with in the UI. We enforce metrics not to be a string to precisely avoid this issue - UIs should render them consistently based on the unit they have associated with.

entityBot (approved):

PR #945, opened 2023-03-01
feat: create synthetics runtime upgrade validation entity

mderemer-nr:
purpose: this will give us a way to index information for the ui to display about a monitors current runtime upgrade validation.
account|SYNTH|RUNTIME_VALIDATION|domainId
domainId is a 32 char uuid.

### Relevant information

### Checklist

 explanation above.

### DIFF
--- a/definitions/synth-runtime_validation/summary_metrics.yml
+++ b/definitions/synth-runtime_validation/summary_metrics.yml
@@ -1,0 +1,20 @@
+monitorName:
+  title: Monitor Name
+  unit: STRING
+  tag:
+    key: monitorName
+monitorType:
+  title: Monitor Type
+  unit: STRING
+  tag:
+    key: monitorType
+validationStatus:
+  title: Validation Status
+  unit: STRING
+  tag:
+    key: validationStatus
+lastValidated:
+  title: Last Validated
+  unit: TIMESTAMP
+  tag:
+    key: lastValidated
--- a/definitions/synth-runtime_validation/summary_metrics.yml
+++ b/definitions/synth-runtime_validation/summary_metrics.yml
@@ -1,8 +1,8 @@
-monitorName:
+name:
   title: Monitor Name
   unit: STRING
   tag:
-    key: monitorName
+    key: name
 monitorType:
   title: Monitor Type
   unit: STRING
--- a/definitions/synth-runtime_validation/summary_metrics.yml
+++ b/definitions/synth-runtime_validation/summary_metrics.yml
@@ -1,8 +1,3 @@
-name:
-  title: Monitor Name
-  unit: STRING
-  tag:
-    key: name
 monitorType:
   title: Monitor Type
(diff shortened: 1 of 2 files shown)


entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

definitions/synth-runtime_validation/summary_metrics.yml
```diff
@@ -0,0 +1,20 @@
+monitorName:
```
codymjarrett, line 1: Could we change this key to `name` instead of `monitorName`. This request is two fold, the `EntityTableRowCell` component expects a `name` field on the entity in order to render correctly. We could do some transforming on the client but if we could fix this on the server, even better. Also, I noticed that the current entities in the current monitor list have a `name` field so this would be consistent with that implementation.

entityBot (approved):

entityBot (approved):

definitions/synth-runtime_validation/summary_metrics.yml
mderemer-nr: updated field but left title the same. Is this alright with you @codymjarrett ?

codymjarrett: Thanks for that. Honestly, I don't know too much about how entities are created here. It's interesting because looking at the [entity definition for the current monitor list](https://github.com/newrelic/entity-definitions/blob/main/definitions/synth-monitor/definition.yml) there isn't a `name` or `monitorName` field defined however there's a `name` field on the entity objects returns from the response. Is there some magic happening that isn't explicitly detailed here?  @mderemer-nr

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

definitions/synth-runtime_validation/summary_metrics.yml
gvassallo: @codymjarrett that is because the entity is not created via synthesis. If it were, then you'd have to define how to map a specific attribute from the telemetry to a specific entity field. 

As this is not the case for this entity, it will be up to the (New Relic internal) system that generates those entities to specify the value for those fields.

entityBot (approved):

entityBot (approved):

gvassallo:
@mderemer-nr @codymjarrett  can I go ahead and merge the PR?

gvassallo (approved):

entityBot (approved):

entityBot (approved):

mderemer-nr:
@gvassallo Yes its ready for merge. Not sure if i have write perms.

entityBot (approved):

entityBot (approved):

PR #960, opened 2023-03-08
feat: add kubernetes ingress entity

philomory:
### Relevant information

This PR adds an entity to represent Kubernetes Ingress objects. Specifically, this synthesizes Ingress objects from the metrics exported by the `ingress-nginx` ingress controller, though it leaves room for the addition of rules to synthesize Ingress objects from the metrics of other ingress controllers in the future.

There _is_ one concern I have when it comes to adding additional rules for other ingress controllers: adding different _synthesis rules_ for different ingress controllers seems straightforward enough, as the `conditions` field would allows you to distinguish between data sources for the same entity via e.g. metrics name prefixes, such as the `nginx_ingress_controller_` prefix vs e.g. the `ambassador_` prefix or the `traefik_` prefix. But the Golden Metrics definitions in `golden_metrics.yml` can only distinguish between data sources for the same entity via the `instrumentation.provider` and/or `instrumentation.name` fields, and for all of these different ingress controllers, those fields would by default be `prometheus` and `remote-write`, respectively (some ingress controllers may optionally be able to export metrics in other formats besides Prometheus, but to the extent that I am familiar with other ingress controllers besides `ingress-nginx`, Prometheus metrics seem to be a pretty common default format).

That said, I think that addressing that particular concern can be handled in a separate issue. See also [this Slack thread](https://newrelicusers.slack.com/archives/C03LSNGL491/p1678306696238199?thread_ts=1678138168.142379&cid=C03LSNGL491).

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-kubernetes_ingress/definition.yml
+++ b/definitions/ext-kubernetes_ingress/definition.yml
@@ -1,0 +1,31 @@
+domain: EXT
+type: KUBERNETES_INGRESS
+
+goldenTags:
+  - cluster_name
+  - k8s.namespace
+  - ingress_class
+
+synthesis:
+  rules:
+    - compositeIdentifier:
+        separator: "/"
+        attributes:
+          - cluster_name
+          - exported_namespace
+          - ingress
+      encodeIdentifierInGUID: true
+      name: ingress
+      conditions:
+        - attribute: metricName
+          prefix: nginx_ingress_controller_
+      tags:
+        k8s.namespace:
+          entityTagNames: [exported_namespace]
+          multiValue: false
+        ingress_class:
+          entityTagNames: [controller_class]
+          multiValue: false
+
+configuration:
+  alertable: true
--- a/definitions/ext-kubernetes_ingress/golden_metrics.yml
+++ b/definitions/ext-kubernetes_ingress/golden_metrics.yml
@@ -1,0 +1,20 @@
+requestsPerSecond:
+  title: Requests per second
+  unit: REQUESTS_PER_SECOND
+  queries:
+    prometheus:
+      select: rate(sum(nginx_ingress_controller_request_size_count), 1 SECONDS) AS 'Requests Per Second'
+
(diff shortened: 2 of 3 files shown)


entityBot (changes requested):
beep boop bop.

 ```
* Error VALIDATING domainType: 'EXT-KUBERNETES_INGRESS' on Golden metrics for key 'SELECT filter( rate(Sum( nginx_ingress_controller_requests),1 seconds),  WHERE status Not RLIKE '[4-5].*') * 100  / filter( rate(sum( nginx_ingress_controller_requests), 1 seconds), WHERE status RLIKE '.*') AS 'Error Rate' FROM Metric WHERE entity.guid IN ('guid') FACET entity.name TIMESERIES' - Exception: The select clause cannot be used to produce a valid golden metric.
```

**Review by entityBot — CHANGES_REQUESTED** [review]
beep boop bop.

 ```
* Error VALIDATING domainType: 'EXT-KUBERNETES_INGRESS' on Golden metrics for key 'SELECT filter( rate(Sum( nginx_ingress_controller_requests),1 seconds),  WHERE status Not RLIKE '[4-5].*') * 100  / filter( rate(sum( nginx_ingress_controller_requests), 1 seconds), WHERE status RLIKE '.*') AS 'Error Rate' FROM Metric WHERE entity.guid IN ('guid') FACET entity.name TIMESERIES' - Exception: The select clause cannot be used to produce a valid golden metric.
```

philomory:
Hm, interesting. When I rand the validations locally it said there were no errors. I'll have to double-check.

otaviocarvalho:
Just for internal reference, there is another team working to integrate ingress k8s entities too, we would need to sync up and converge both: https://newrelic.slack.com/archives/CH5MLECNN/p1680022972234989?thread_ts=1679906231.458659&cid=CH5MLECNN

naxhh:
Pr has not moved in 1 month, closing.

PR #973, opened 2023-03-15
FEAT: Add Gigamon sourced DHCP Network Service

nr-jbiggley:
### Relevant information

Describe what you have done and any details that you think are relevant or that you might want to discuss with us. 

### Checklist

* [X] I've read the guidelines and understand the acceptance criteria.
* [X] The value of the attribute marked as `identifier` will be unique and valid. 
* [X] I've confirmed that my entity type wasn't already defined. If it is I'm providing an
 explanation above.

PR #980, opened 2023-03-23
Validator fixes on equality rule check and tone it down logs

naxhh:
### Relevant information

Remove logs that only said "<DOMAIN-TYPE> valid" and only log the actual errors.
This will make it much easier to check the CI results.

In https://github.com/newrelic/entity-definitions/pull/973 we noticed that the validator was failing but the conditions are not the same.

After some checks, we realized that the way how the conditions are checked was wrong.
Before it triggered when some condition was equal.

Now it sorts both sets and compares them, if all of them are equal then we fail the check.

### Checklist

 explanation above.


entityBot (approved):

entityBot (approved):

entityBot (approved):

otaviocarvalho (approved):

validator/tools/validate_rules.js
```diff
@@ -141,7 +155,6 @@ RULES.forEach(rule => {
                         // terminate early
                         process.exit(1)
                     }
-            console.log(`=> processed ${definition.domain}-${definition.type}... valid for rule "${rule.name}"`)
```
otaviocarvalho, line 144: Danke!

entityBot (approved):

entityBot (approved):

andre-nr (approved):

PR #995, opened 2023-03-29
Remove unused Kubernetes entities synthesis definitions

htroisi:
### Relevant information

This PR removes the synthesis rules from each Kubernetes entity's definition file:

- These synthesis rules are not actually used, since Kubernetes entities are synthesized using hardcoded rules in the `entity-synthesis-helper` repo.
- These synthesis rules were added during the initial stages of a project to migrate Kubernetes entity synthesis from `entity-synthesis-helper`  to the entity platform. However, this project has been put on hold. 
- Until the migration project becomes a priority again, we should remove these synthesis definitions to prevent any confusion. @naxhh

### Checklist

 explanation above.


entityBot (approved):

entityBot (approved):

naxhh (approved):
Let's check that we are actually not creating anything from these synthesis rules.

It would not be the first time we are surprised on a rule being used that was here just for test

otaviocarvalho (approved):
Thanks for the clean up! It will save us from some confusion during incidents in the future

entityBot (approved):

entityBot (approved):

PR #1007, opened 2023-04-10
add service.domain for opentelemetry tags

loomis-relativity:
### Relevant information

The organization I work for uses OpenTelemetry for formatting/transmitting service telemetry and Backstage to catalog services. The current `service.name` and `service.namespace` map well to the Backstage entity and system objects. We also want to capture the Backstage domain, which groups the `service.namespace` (system) values. This change will tag entities with `service.domain`, if it is defined in the telemetry payload. 

### Checklist

 explanation above.

### DIFF
--- a/definitions/ext-service/definition.yml
+++ b/definitions/ext-service/definition.yml
@@ -54,6 +54,7 @@ synthesis:
         entityTagName: language
       telemetry.sdk.version:
       service.namespace:
+      service.domain:
 
     # telemetry from opentelemetry provider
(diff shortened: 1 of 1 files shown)


entityBot (approved):

entityBot (approved):

loomis-relativity:
Used the `tags.` prefix to attach this attribute to entities. This is unlikely to be standardized based on the move toward the Elastic Common Schema.

PR #1098, opened 2023-05-30
[NO-JIRA] chore: update NODE docker version to avoid security issues

manuelrodriguez-nr:
### Checklist


entityBot (approved):

entityBot (approved):

naxhh (approved):
If you can upgrade https://github.com/newrelic/entity-definitions/blob/main/.nvmrc so local dev uses the same version that would be nice

entityBot (approved):

entityBot (approved):

PR #1130, opened 2023-06-28
[Product Partnerships] Added AWS DataSync

rahul188:
### Checklist

### DIFF
--- a/definitions/infra-awsdatasync/golden_metrics.yml
+++ b/definitions/infra-awsdatasync/golden_metrics.yml
@@ -1,0 +1,18 @@
+bytesWritten:
+  title: Bytes written
+  unit: COUNT
+  queries:
+    aws:
+      select: sum(aws.datasync.BytesWritten)
+      from: Metric
+      eventId: entity.guid
+      eventName: entity.name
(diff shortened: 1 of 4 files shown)


entityBot (approved):

entityBot (approved):

definitions/infra-awsdatasync/golden_metrics.yml
```diff
@@ -0,0 +1,18 @@
+bytesWritten:
+  title: Bytes written
+  unit: COUNT
+  queries:
+    aws:
```
gvassallo, line 5: @rahul188 are these metrics reported as Events as well? If so, you'll have to provide the event equivalent GM definition here as well. 
[Example](https://github.com/newrelic/entity-definitions/blob/main/definitions/infra-awsalb/golden_metrics.yml#L10)

rahul188: @gvassallo, i feel we are safe to move ahead with bytesWritten & filesTransferred Metrics.

gvassallo (approved):

PR #1142, opened 2023-07-17
Proposal of new documentation structure

otaviocarvalho:
### Relevant information

The current documentation is mainly focused on entities (since the repository was primarily created for its definitions). 

However, since we are now expanding to include relationship definitions too, I would to suggest a more normalized structure for our documentation.

Proposals:
1. Keep the main README.md (and further expand it on the [next PR](https://github.com/newrelic/entity-definitions/pull/1138/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R18) to point to the respective entities and relationship docs)
2. Keep a `/docs` repository containing the detailed documentation
3. Create two sub-folders to group the documentation `/docs/entities` and `/docs/relationships` with their respective README.md
4. Create a README.md inside `/docs` pointing to entities/relationships respective README(s)

Screenshot 2023-07-17 at 16 12 17

### Checklist


entityBot (approved):

otaviocarvalho:
Didn't switch the folder name from `/definitions` to `/entities` (we will have a new `/relationships`) because it may break the validators, but it is something we could consider too. Wdyt?

entityBot (approved):

entityBot (approved):

naxhh:
> Didn't switch the folder name from /definitions to /entities (
This is something we must do IMHO, I proposed entityTypes in the relationship docs but I also didn't want to mix one work with the other.

IMO we should first make the emitter work with both folder names (old & new) then merge as much PR as possible to make the transition easier, and finally rename the folder.

If you want do it as part of the current work is fine with me, we can also open a ticket for it and do it later.

naxhh (approved):

apoloa (approved):
💯

entityBot (approved):

entityBot (approved):

PR #1165, opened 2023-08-03
APM calls REDISINSTANCE: Propose a change to the rule to use APM metrics

naxhh:
### Relevant information

Use APM metrics instead of DT.

Users don't need to opt-in into APM metrics and hence we will create more relationships than using DT for this.

Note: Approval is left to the team proposing these new relationships. Don't merge until they approve it.

### Checklist


entityBot (approved):

entityBot (approved):

naxhh:
internal conversation in https://newrelic.slack.com/archives/C05F5RU48NL/p1690991677333779

entityBot (approved):

entityBot (approved):

naxhh:
@otaviocarvalho merging this one so it goes along with the regex fix.

It makes sure we do the lookup and use apm metrics,

PR #1180, opened 2023-08-10
Add apm to apm relationship by kafka topic

durgaprasad95:


javimb:
Thanks for contributing @durgaprasad95!
It would be helpful if you add some context about this change. What's the purpose of this relationship that you're introducing? It seems to connect an APM application with itself if it reports the `MessageBroker/Kafka/Internal` metric. Why is this important to you?

By the way, we're currently working on adding relationships between APM applications and Kafka topics. Is that something useful for you? Would it be aligned with your needs?

naxhh:
Closing old issue without activity

PR #1254, opened 2023-10-04
Updating prometheus golden metric definitions

bonczj:
### Relevant information
This PR is to update several prometheus golden metrics so that these metrics will actually calculate values for customers.

Prometheus CPU usage golden metric had a condition where the value was used in a way that could lead to a divide by zero instance. Added a where clause to prevent this from occurring.

Prometheus memory usage and storage usage both needed to utilize two different metrics to compute the value. We have created a script that will precompute these values on the host and can be sent in via the prometheus node_exporter using a textfile collector. With this, both metrics can be simplified to a single value.

### Checklist

* [ x] I've read the guidelines and understand the acceptance criteria.


entityBot (approved):

bonczj:
Closing and will reopen under a different email address so CLAassistant is no longer confused.

PR #1271, opened 2023-10-20
Use the convert function for newrelic-opentelemetry data

sdaubin:
### Relevant information

This updates one golden metrics query to use the new NRDB `convert` function to convert the measurement to `ms` instead of assuming the unit is in `s`.  This only applies when the `instrumentation.provider` is `newrelic.opentelemetry`.  We do not have any production agents reporting this configuration.  We are started to alpha test this with a small handful of customers.  The risk is very low.

Before deploying this change to production, we need to verify that the NRDB team has enabled the `convert` function in the production environment.

https://docs.google.com/document/d/1nuRiK9HM-x2d52mtMLuiPUkSUzONfKFtTFFmRpOocoU/edit?usp=sharing

### Checklist

### DIFF
--- a/entity-types/ext-service/golden_metrics.yml
+++ b/entity-types/ext-service/golden_metrics.yml
@@ -47,7 +47,7 @@ responseTimeMs:
       from: Span
       where: span.kind LIKE 'server' OR span.kind LIKE 'consumer' OR kind LIKE 'server' OR kind LIKE 'consumer'
     newrelic-opentelemetry:
-      select: average(apm.service.transaction.duration * 1000)
+      select: average(convert(apm.service.transaction.duration, unit, 'ms'))
       from: Metric
     kamon-agent:
       select: average(http.server.requests)


entityBot (approved):

entityBot (approved):

lujop:
LGTM @sdaubin 
But let's wait for confirmation that this new function is supported in PROD by Golden Signals/Collector before merging it.

lujop (approved):
LGTM and I will merge it as I can use the convert function in PROD, and also I've received confirmation from the Golden Metrics and Collector team that they support it.

entityBot (approved):

entityBot (approved):

sdaubin:
@lujop The pipelines should be updated to support the function, but we need the nrdb team to enable it in production (it's enabled in staging).  Were you able to get confirmation from them?

PR #1295, opened 2023-12-08
Feat: SentrySoftware entity creation

jcountsNR:
### Checklist


jcountsNR:
I'm not sure why my fork included all of those old commits, but the only changes should be from the most recent commit, 2 new files.

lujop:
@jcountsNR Could you start a new PR and apply only the relevant commits?

apoloa:
Don't worry for the commits we will squash them when we will merge the pr.

entity-types/ext-sentrysoftware/dashboard.json
```diff
@@ -0,0 +1,1124 @@
+{
+    "name": "Hardware Sentry",
```
apoloa, line 2: Can you set the same name as the entity one?

entity-types/ext-sentrysoftware/definition.yml
```diff
@@ -0,0 +1,24 @@
+domain: EXT
+type: SENTRYSOFTWARE
```
apoloa, line 2:
```suggestion
type: SENTRY_SOFTWARE
```

entity-types/ext-sentrysoftware/definition.yml
```diff
@@ -0,0 +1,24 @@
...
+synthesis:
+  rules:
+  - identifier: agent.host.name
```
apoloa, line 5: Are you sure that the name will be enough to identify the entity? Let me set an example: If the account contains two sentry software with the same `agent.host.name`, it will show only one entity in New Relic.

apoloa (changes requested):

entityBot (approved):

entityBot (approved):

PR #1301, opened 2023-12-12
[Product Partnerships] Added AWS VPN Entity

rahul188:
### Checklist

### DIFF
--- a/entity-types/infra-awsvpn/summary_metrics.yml
+++ b/entity-types/infra-awsvpn/summary_metrics.yml
@@ -1,0 +1,18 @@
+providerAccountName:
+  title: AWS account
+  unit: STRING
+  tag:
+    key: providerAccountName
+region:
+  title: Region
+  unit: STRING
+  tag:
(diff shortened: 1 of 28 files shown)


entity-types/infra-awsvpn/summary_metrics.yml
```diff
@@ -0,0 +1,18 @@
...
+  tag:
+    key: aws.awsRegion
+tunnelDataIn:
+  goldenMetric: tunnelDataIn
+  title: Tunnel data in
+  unit: COUNT
+tunnelDataOut:
+  goldenMetric: tunnelDataOut
+  title: Tunnel data Out
+  unit: COUNT
```
lujop, line 18: NIT: Would it be better to use some more precise unit like https://github.com/newrelic/entity-definitions/blob/main/docs/entities/golden_metrics.md#metric-unit (also for the dashboard)

lujop (approved):
LGTM, I will just change the COUNT unit if a better one is applicable

PR #1313, opened 2023-12-15
Convert elasticsearch node dashboard back to sample

jlegoff:
### Relevant information

Convert elasticsearch node dashboard back to sample

### Checklist


naxhh:
@entityBot

PR #1427, opened 2024-01-29
Adding Aws Arn tag for dynamoDB relationships

DasuNewrelic:
### Relevant information

1. Currently we are working with mapping relationships for Lambda functions for Otel entity. We've already added new rules for DynamoDB calls. Also adding one more field i.e aws.arn for the same.
2. Please review this PR & provide the feedback on it, And also guid me for further assistance.
3. Please find the ticket link [here](https://new-relic.atlassian.net/browse/NR-216368)

### Checklist

### DIFF
--- a/relationships/candidates/AWSDYNAMODBTABLE.yml
+++ b/relationships/candidates/AWSDYNAMODBTABLE.yml
@@ -12,6 +12,8 @@ lookups:
           field: awsRegion
         - tagKeys: ["aws.tableName"]
           field: awsTableName
+        - tagKeys: ["aws.arn"]
+          field: awsArn
     onMatch:
       onMultipleMatches: RELATE_ALL
     onMiss:
--- a/relationships/candidates/AWSDYNAMODBTABLE.yml
+++ b/relationships/candidates/AWSDYNAMODBTABLE.yml
@@ -13,7 +13,7 @@ lookups:
         - tagKeys: ["aws.tableName"]
           field: awsTableName
         - tagKeys: ["aws.arn"]
-          field: awsArn
+          field: cloudArn
     onMatch:
       onMultipleMatches: RELATE_ALL
     onMiss:
--- a/relationships/candidates/AWSDYNAMODBTABLE.yml
+++ b/relationships/candidates/AWSDYNAMODBTABLE.yml
@@ -6,14 +6,8 @@ lookups:
     tags:
       matchingMode: ALL
       predicates:
-        - tagKeys: ["aws.accountId"]
-          field: awsAccountId
-        - tagKeys: ["aws.awsRegion"]
-          field: awsRegion
-        - tagKeys: ["aws.tableName"]
-          field: awsTableName
         - tagKeys: ["aws.arn"]
-          field: cloudArn
+          field: cloudResourceId
     onMatch:
       onMultipleMatches: RELATE_ALL
     onMiss:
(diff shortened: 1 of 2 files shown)


entityBot (approved):

entityBot (approved):

entityBot (approved):

relationships/candidates/AWSDYNAMODBTABLE.yml
```diff
@@ -12,6 +12,8 @@ lookups:
           field: awsRegion
         - tagKeys: ["aws.tableName"]
           field: awsTableName
+        - tagKeys: ["aws.arn"]
```
javimb, line 15: This is redundant and should be move to another rule.
Entity should match awsAccountId AND awsRegion AND awsTableName, OR arn.

relationships/candidates/AWSDYNAMODBTABLE.yml
```diff
@@ -12,6 +12,8 @@ lookups:
...
+          field: cloudArn
```
javimb, line 16: I think that the term `cloudArn` is confusing. ARN is an AWS concept, so it doesn't make sense to generalize it as a cloud concept. I'd choose something more specific like `awsArn` or something more generic like `cloudResourceId`.

javimb (changes requested):

relationships/candidates/AWSDYNAMODBTABLE.yml
DasuNewrelic: Can you please suggest that how I should change.?

javimb: You could add a different `lookup` rule for ARN as we do in the candidate definition for `HOST`: https://github.com/newrelic/entity-definitions/blob/main/relationships/candidates/HOST.yml

entityBot (approved):

entityBot (approved):

relationships/candidates/AWSDYNAMODBTABLE.yml
DasuNewrelic: I've just added aws.arn and other attributes are removed. Can please have a look.

javimb (approved):

relationships/candidates/AWSDYNAMODBTABLE.yml
naxhh: is intended that account, region & table are nowhere in any rule now?

manuelrodriguez-nr: arn contains all that, right? [LINK](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html)
```
arn:partition:service:region:account-id:resource-id
arn:partition:service:region:account-id:resource-type/resource-id
arn:partition:service:region:account-id:resource-type:resource-id
``` 
so what I understand is that this rule is for ARN only and there will be another one for accountId+Region+TableName. Is this correct @DasuNewrelic ?

DasuNewrelic: Yes, arn contains all.

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

PR #1434, opened 2024-02-05
feat: APM-APPLICATION to INFRA-AWSMSKTOPIC relationship rule

durgaprasad95:
### Relevant information

1. Rename the relationship rule for consistency with the rest of the rules that are related to INFRA .
2. Use the proper candidate for the synthesis lookup

### Checklist


relationships/candidates/AWSMSKTOPIC2.yml
```diff
@@ -0,0 +1,20 @@
+category: AWSMSKTOPIC2
```
javimb, line 1: Is there any reason to create a new category rather than updating `AWSMSKTOPIC`?
`AWSMSKTOPIC` category is not used in any relationship rule, so it shouldn't be a problem.

naxhh:
closing as stale. Feel free to re-open once the feedback has been addressed.

PR #1468, opened 2024-02-15
feat: Add renaming for golden tags in OS K8s entities

brettj-newrelic:
### Relevant information
A tag remapping for golden tag on OS instrumented K8s workloads, persistentvolume, pvc.

### Checklist

### DIFF
--- a/entity-types/infra-kubernetes_cronjob/definition.yml
+++ b/entity-types/infra-kubernetes_cronjob/definition.yml
@@ -42,3 +42,10 @@ synthesis:
         # value added for test entities only
         - attribute: newrelicOnly
           value: "true"
+      tags:
+        cronjob:
+          entityTagNames: [k8s.cronjobName]
+        namespace:
+          entityTagNames: [k8s.namespaceName]
+        k8s.cluster.name:
+          entityTagNames: [k8s.clusterName]
--- a/entity-types/infra-kubernetes_persistentvolumeclaim/definition.yml
+++ b/entity-types/infra-kubernetes_persistentvolumeclaim/definition.yml
@@ -42,3 +42,10 @@ synthesis:
         # value added for test entities only
         - attribute: newrelicOnly
           value: "true"
+      tags:
+        persistentvolumeclaim:
+          entityTagNames: [k8s.pvcName]
+        namespace:
+          entityTagNames: [k8s.pvcNamespace]
+        k8s.cluster.name:
+          entityTagNames: [k8s.clusterName]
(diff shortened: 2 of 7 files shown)


entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

rodolfocal (approved):

entity-types/infra-kubernetes_cronjob/definition.yml
```diff
@@ -41,4 +41,11 @@ synthesis:
...
-          value: "true"
\ No newline at end of file
+          value: "true"
+      tags:
+        k8s.cronjobName:
+          entityTagNames: [cronjob]
+        k8s.namespaceName:
+          entityTagNames: [namespace]
+        k8s.clusterName:
+          entityTagNames: [k8s.cluster.name]
```
rodolfocal, line 51: Why not use the same "tag names" that come from telemetry. Users tend to be confused when the names on telemetry and on tags are different from each other.

On the same matter, the golden tags definition have different tag names on them: https://github.com/newrelic/entity-definitions/pull/1468/files#diff-0c1214b8d257118b55b64d763c1e5d1a0e05b73da6c741e81f9f6f284d115964R11-R13

entityBot (approved):

entityBot (approved):

entity-types/infra-kubernetes_cronjob/definition.yml
brettj-newrelic: I misunderstood the tag names and attribute names worked here, so the attribute sources and tag name were flipped. This should be resolved with the updates.

entity-types/infra-kubernetes_persistentvolumeclaim/definition.yml
```diff
@@ -41,4 +41,11 @@ synthesis:
...
         # value added for test entities only
         - attribute: newrelicOnly
-          value: "true"
\ No newline at end of file
+          value: "true"
+      tags:
+        persistentvolumeclaim:
+          entityTagNames: [k8s.pvcName]
+        namespace:
+          entityTagNames: [k8s.pvcNamespace]
```
javimb, line 49: Namespace tag is consistently named as `k8s.namespace` in all other entities, but here it contains the `pvc` prefix. Shouldn't we remove the prefix?

brettj-newrelic: It agree that `k8s.namespace` makes more sense – but `k8s.pvcNamespace` is the existing golden tag for `perisistentvolumeclaim` – see line 13.

entityBot (approved):

javimb (approved):

PR #1584, opened 2024-04-10
Fix OTEL host service relationship

jlegoff:
### Relevant information

OTEL hosts' entity guids were recently changed ([PR](https://github.com/newrelic/entity-definitions/pull/1499)). This PR adapts the service-to-host relationship to use the correct GUID.

What's more, an old rule for the OTEL collector was deleted, as it's no longer relevant.

### Checklist

### DIFF
--- a/relationships/synthesis/INFRA-HOST-to-EXT-SERVICE.yml
+++ b/relationships/synthesis/INFRA-HOST-to-EXT-SERVICE.yml
@@ -29,36 +29,3 @@ relationships:
           attribute: entity.guid
           entityType:
             attribute: entity.type
-
-  - name: otelServiceHostMetric
-    version: "1"
-    origins:
-      - OpenTelemetry
-    conditions:
-      - attribute: eventType
-        anyOf: [ "Metric" ]
-      - attribute: newrelic.source
-        anyOf: [ "api.metrics.otlp" ]
-      - attribute: service.name
-        anyOf: [ "io.opentelemetry.collector" ]
-    relationship:
-      expires: P75M
-      relationshipType: HOSTS
-      source:
-        buildGuid:
-          account:
-            attribute: accountId
-          domain:
-            value: INFRA
-          type:
-            # This is an otel host hence no use of NA on the type.
-            value: HOST
-          identifier:
-            fragments:
-              - attribute: host.id
-            hashAlgorithm: FARM_HASH
-      target:
-        extractGuid:
-          attribute: entity.guid
-          entityType:
-            attribute: entity.type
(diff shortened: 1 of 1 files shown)


relationships/synthesis/INFRA-HOST-to-EXT-SERVICE.yml
```diff
@@ -28,37 +28,4 @@ relationships:
...
-    version: "1"
-    origins:
-      - OpenTelemetry
-    conditions:
-      - attribute: eventType
-        anyOf: [ "Metric" ]
-      - attribute: newrelic.source
-        anyOf: [ "api.metrics.otlp" ]
-      - attribute: service.name
-        anyOf: [ "io.opentelemetry.collector" ]
```
alanwest, line 43: Maybe it makes sense to keep this rule and remove the `service.name` condition. It may also make sense to add a rule for logs. That is, an OTel service that includes a `host.id` attribute on its telemetry, will normally include it on its spans, metrics, and logs that it emits.

entityBot (approved):

entityBot (approved):

relationships/synthesis/INFRA-HOST-to-EXT-SERVICE.yml
```diff
@@ -18,8 +20,8 @@ relationships:
           domain:
             value: INFRA
           type:
-            # This is an otel host hence no use of NA on the type.
```
jack-berg, line 21: @jlegoff we want the relationship to be created whether the host entity is synthesized from data coming from _either_ the collector w/ host metrics receiver OR the new relic infra agent.

Will these rules accommodate both cases?

jlegoff: Good question! I didn't know about this other use case.

In theory nothing prevents this rule from creating a relationship with the infra agent _provided_ the `host.id` attribute sent by the service matches the host's domain id.

This, however, might prove cumbersome for customers. OTEL's [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/host.md) state that `host.id` should be:
* the instance id for cloud entities
* the machine id for non-containerized systems

I don't know whether the conventions for the infra agent are documented. It's something like:
* if the `display_name` parameter is set in the agent config, then use this
* _else if_ it's a cloud instance, then use the instance id
* _else_ use the fqdn

So as a customer, you will need to know which `host.id` to set in order for the relationship to be synthesised. This should be properly documented.

There is another caveat, which is that infra host entities used to have a random domain id assigned to them. This is not the case anymore, but there are still some infra hosts in production for which the relationships won't be created.

To sum it up:
* this rule allows relationships to be created with NR hosts
* we need to document and test how exactly to achieve this

entityBot (approved):

entityBot (approved):

alanwest (approved):

andre-nr (approved):

PR #1616, opened 2024-05-02
feat: add VSPHEREVM to HOST relationship

sigilioso:
### Relevant information

This PR introduced a relationship between `VSPHEREVM` entities and `HOST` entities which was requested by a customer.

The vCenter/vSphere setup setup can be complex and very different depending on the user needs, and depending on how hostnames are defined in the customer's vCenter, this relationship may or may not be suitable. This is the reason why an additional condition has been introduced: the relationship will be created only if the corresponding telemetry contains the attribute `label.enableVMHostNameRelationship`.

The PR will remain as Draft, until clarifying some doubts:

- [x] ~Is `CONTAINS` the proper/expected relationship type? (To be clarified with the PM)~ 
  - Confirmed, `CONTAINS` is the expected relationship type.
- [x] ~The relationship definition uses a candidate relationship for HOST that already existed. Specifically [HOST category](https://github.com/sigilioso/entity-definitions/blob/4544fed9d5d0ff08c0fd005308a0820c7dc09a43/relationships/candidates/HOST.yml#L1). We need to clarify if the existing lookup with `matchingMode: FIRST` will work even if only one of the fields is present.~. 
  - Since no uninstrumented entities are expected, we will probably need a different candidate.
- [x] ~If the the candidate relationship specified above worked, would its current definition suit the expectations? If no HOST entity is found an uninstrumented entity will be created. (To be clarified with the PM).~ 
  - It doesn't suit the expectations, for this particular scenario no unistrumented entity should be created.

### Checklist


entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

relationships/candidates/HOSTNAMEHOST.yml
```diff
@@ -0,0 +1,14 @@
+category: HOSTNAMEHOST
```
sigilioso, line 1: Even if there is another category (`HOST`) pointing to the INFRA-HOST entity, a new category was created because the existing one includes a lookup entry which would also match by `hostname` and it would create uninstrumented entities on miss.

entityBot (approved):

entityBot (approved):

relationships/candidates/HOSTNAMEHOST.yml
EmiliaRipeanu: I think you can use the HOST category but change the hostname field to something else like onhostEntityHostname

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

relationships/candidates/HOSTNAMEHOST.yml
sigilioso: Thanks! I didn't realize it is possible to use a different field name. I've updated it here: ff94074, let me know what you think.

relationships/synthesis/INFRA-VSPHEREVM-to-INFRA-HOST.yml
```diff
@@ -0,0 +1,30 @@
...
+    conditions:
+      - attribute: eventType
+        anyOf: ["VSphereVmSample"]
+      - attribute: vmHostname
+        present: true
+      # This additional attribute is required because the relationship doesn't apply to every environment
+      # (the VSPHEREVM vmHostname and the HOST hostname don't match in every scenario).
+      # Adding this label to VSphereVmSamples means that a relationship between the VSPHEREVM entity and the
+      # HOST entity whose hostname matches the vmHostname is expected.
+      - attribute: label.enableVMHostNameRelationship
```
EmiliaRipeanu, line 15: Can we also add additional information about this attribute? For example, we might want to explain the billing implications in case other customers see this rule and decide to implement this hack. The one for which we're adding the rule now knows about this but other don't.

sigilioso: I think it makes total sense to document those implications. But I'd say that should be documented along with the instructions to _enable_ this relationship (probably in the [vsphere integration documentation](https://docs.newrelic.com/docs/infrastructure/host-integrations/host-integrations-list/vmware-vsphere-monitoring-integration/). After all, the label will be added in the integration configuration. Besides, adding this label is not different from adding any other [arbitrary label](https://docs.newrelic.com/docs/infrastructure/host-integrations/infrastructure-integrations-sdk/specifications/host-integrations-standard-configuration-format/#labels). WDYT?

entityBot (approved):

entityBot (approved):

entityBot (approved):

andre-nr (approved):

PR #1662, opened 2024-06-13
Experimental rule to synthesize APM-APPLICATION entity from OpenTelemetry

alanwest:
OpenTelemetry instrumented applications are currently synthesized as `EXT-SERVICE` entities. We want to experiment with synthesizing them as `APM-APPLICATION` entities.

This rule is copied from the following rule https://github.com/newrelic/entity-definitions/blob/8565a63efbacd3ded986dd3271543021b60d3dd7/entity-types/ext-service/definition.yml#L40-L62.

An additional condition has been added `newrelic.entity.type = apm`. The intent is to allow us to experiment with things to determine what breaks but without affecting anyone else.

@tkarai Ethan mentioned that he was speaking with you about this. Can you confirm this rule will achieve what we want?


entityBot (approved):

entityBot (approved):

javimb:
@alanwest PR looks good in general, but since it's a change that could have some implications with the `EXT-SERVICE` synthesis rules, I'd recommend to [add test](https://github.com/newrelic/entity-definitions?tab=readme-ov-file#how-to-add-testing-data).

I think it would be a good idea to a some tests for `EXT-SERVICE` first. The validator would check if this new definition would be incorrectly triggered by them. Then we can add test for this definition.

tkarai:
> @tkarai Ethan mentioned that he was speaking with you about this. Can you confirm this rule will achieve what we want?

I believe it's worth double-checking with Entity Platform folks in their help channel.

entityBot (approved):

alanwest:
@javimb I've added some tests.

@tkarai I've posted [a question](https://newrelic.slack.com/archives/CH5MLECNN/p1721239051478419) in #help-entity-platform

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

manuelrodriguez-nr:
@entityBot  deploy

naxhh (approved):
after some talks we will merge this one to ease testing

PR #1671, opened 2024-06-21
NR-259488-01: Including Primary endpoint and reader endpoints in the candidate file

AravindKumar8520:
### Including Primary endpoint and reader endpoints in the candidate file

According to the cluster configurations, If the customer has selected the cluster mode as enabled while create a redis server , then there will be a configuration endpoint; if the customer has selected the cluster mode as disabled, they will have either a primary or reader endpoint. The PR allows these three types of endpoints when connecting to the cluster/node.

### Checklist


entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

relationships/candidates/REDIS.yml
```diff
@@ -8,9 +8,9 @@ lookups:
     tags:
       matchingMode: ALL
       predicates:
-        - tagKeys: ["aws.elasticache.endpointAddress", "aws.elasticache.configurationEndpointAddress"]
+        - tagKeys: ["aws.elasticache.endpointAddress", "aws.elasticache.configurationEndpointAddress", "aws.elasticache.primaryEndpointAddress", "aws.elasticache.readerEndpointAddress" ]
```
andre-nr, line 11: Does `matchingMode: ALL` make sense in this context or would it need to be `matchingMode: ANY` ?
([docs](https://github.com/newrelic/entity-definitions/blob/main/docs/relationships/candidate_relationships.md#tags-mode))

entityBot (approved):

entityBot (approved):

relationships/candidates/REDIS.yml
AravindKumar8520: I guess, the matching mode refers to the predicates. the endpoint and port, it should match all.

PR #1723, opened 2024-08-07
Stop classifying INFRA-WIN_SERVICE data as EXT-SERVICE

jack-berg:
See https://new-relic.atlassian.net/browse/NR-285288.

cc @alanwest

### DIFF
--- a/entity-types/ext-service/definition.yml
+++ b/entity-types/ext-service/definition.yml
@@ -14,6 +14,13 @@ synthesis:
     - attribute: service_name
     - attribute: newrelic.entity.type
       present: false
+    # Exclude data from INFRA-WIN_SERVICE telemetry sources. Typically, data from these sources
+    # includes entity info (guid, type, name) from the client side. But due to race conditions,
+    # data is occasionally emitted without entity info and is erroneously classified as EXT-SERVICE.
+    # Ideally, we would skip data where collector.name = infrastructure-agent, but it's not possible
+    # to express that condition. Excluding all data with attribute collector.name is probably fine.
+    - attribute: collector.name
+      present: false
     tags:
       telemetry.sdk.name:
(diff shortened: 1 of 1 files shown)


entityBot (approved):

entityBot (approved):

alanwest (approved):
It's unfortunate that we do not know the history of why the `service_name` and `serviceName` rules exist. They certainly have nothing to do with OpenTelemetry, so @jack-berg and myself would not know the history.

I guess there's a small risk here that there exists entities today that are correctly classified as EXT-SERVICE that use `service_name` and also `collector.name`. Though this would be strange since they would unlikely light up any UI experience.

Can anyone from the entity platform team provide any insight into the extent that `service_name` and `serviceName` are used? It'd be great if we could find a way to delete them if possible.

otaviocarvalho:
> It's unfortunate that we do not know the history of why the `service_name` and `serviceName` rules exist. They certainly have nothing to do with OpenTelemetry, so @jack-berg and myself would not know the history.
> 
> I guess there's a small risk here that there exists entities today that are correctly classified as EXT-SERVICE that use `service_name` and also `collector.name`. Though this would be strange since they would unlikely light up any UI experience.
> 
> Can anyone from the entity platform team provide any insight into the extent that `service_name` and `serviceName` are used? It'd be great if we could find a way to delete them if possible.

In summary, I believe it is because it was one of the first rules we created (which predates this public repository) and we were trying to match data points for distinct integrations (e.g. logging used `service.name`, infra used serviceName, etc).

I am not sure if those integrations are using new rules now or they keep relying on what is here.

A bit of code archeology this one to be fairly honest. I have send you some historical data via Slack, hope it helps!

entityBot (approved):

entityBot (approved):

PR #1789, opened 2024-11-19
Entity definition for Media streaming video type entity

jpradhan-nr:
### Relevant information

As media streaming video agent would be ingesting `Video*` event types, We have added this definition for curating media streaming of type video entity from the telemetry being ingested by the media-streaming agents. 

### Checklist

### DIFF
--- a/entity-types/media_streaming-video/definition.yml
+++ b/entity-types/media_streaming-video/definition.yml
@@ -25,4 +25,4 @@ goldenTags:
   - deviceModel
 configuration:
   alertable: true
-  entityExpirationTime: QUARTERLY
+  entityExpirationTime: EIGHT_DAYS
--- a/entity-types/media_streaming-video/definition.yml
+++ b/entity-types/media_streaming-video/definition.yml
@@ -1,0 +1,28 @@
+domain: MEDIA_STREAMING
+type: VIDEO
+synthesis:
+  rules:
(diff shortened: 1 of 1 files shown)


entityBot (approved):

entityBot (approved):

entity-types/media_streaming-video/definition.yml
```diff
@@ -0,0 +1,28 @@
...
+goldenTags:
+  - playerName
+  - playerVersion
+  - deviceType
+  - deviceGroup
+  - deviceManufacturer
+  - deviceModel
+configuration:
+  alertable: true
+  entityExpirationTime: QUARTERLY
```
otaviocarvalho, line 28: Please let's not default to a months old expiration. Entities should be refreshed or keep alive periodically instead. 

Could we please move this one to a lower expiration?

entityBot (approved):

entityBot (approved):

entity-types/media_streaming-video/definition.yml
jpradhan-nr: Thank you, I have changed to `EIGHT_DAYS` which is default.

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

PR #1839, opened 2025-01-06
feat: NR 328461 add CLS browser golden signals

shivalagisetty:
### Relevant information

The Browser is extending the support of CLS (Cumulative Layout Shift ) a core web vital which is missing as the other two are already added. 

Ticket : [https://new-relic.atlassian.net/browse/NR-328461](https://new-relic.atlassian.net/browse/NR-328461)

### Checklist


otaviocarvalho:
We cannot arbitrarily add units on our side. Those need to be changed/allowed on the upstream systems before we can change our validations.

Added more context here: [link](https://newrelic.slack.com/archives/CH5MLECNN/p1736772087561449?thread_ts=1736238264.524199&cid=CH5MLECNN)

entity-types/browser-application/golden_metrics.yml
```diff
@@ -22,6 +22,14 @@ interactionToNextPaint75PercentileS:
     from: PageViewTiming
     eventId: entityGuid
     eventName: appName
+cumulativeLayoutShift75Percentile:
+  title: Cumulative layout shift (75 percentile)
+  unit: STRING
```
vetusbs, line 27: I don't think STRING is the proper unit here.

Is there any specific reason why this should be an STRING?

I totally understand that any other values in the ENUM do not fit completely your needs, but I'd not put an STRING here. 

In many other places, we use just `COUNT`, which is the default.
The `COUNT` unit, is shown as no just a number in the UI.

Let me share some examples:

https://github.com/newrelic/entity-definitions/blob/1ce56fee1df3f75dacfac5c077e2a5711f3cb10d/entity-types/infra-container/summary_metrics.yml#L6

https://github.com/newrelic/entity-definitions/blob/1ce56fee1df3f75dacfac5c077e2a5711f3cb10d/entity-types/infra-awslambdafunctionalias/golden_metrics.yml#L46

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

PR #1856, opened 2025-01-14
Add displayName as a possible tag for k8s candidate matching

atrupiano:
### Relevant information

Adding displayName as a possible tag for matching k8s candidate relationships. We expect displayName to match in the majority of cases, which is why it is being checked before k8s.cluster.name.

### Checklist

* [N/A] The value of the attribute marked as `identifier` will be unique and valid. 
* [N/A] I've confirmed that my entity type wasn't already defined. If it is I'm providing an explanation above.


entityBot (approved):

entityBot (approved):

relationships/candidates/KUBERNETESCLUSTER.yml
```diff
@@ -4,8 +4,10 @@ lookups:
     - domain: INFRA
       type: KUBERNETESCLUSTER
     tags:
-      matchingMode: ANY
+      matchingMode: FIRST
```
otaviocarvalho, line 7: We are changing the matching mode and it is not present on the description of the PR. 

Could we please include the reasoning for this change on the PR description?

otaviocarvalho (changes requested):
Could we please include the reasoning for the matching mode change on the PR description?

otaviocarvalho (approved):

PR #1895, opened 2025-02-03
Add ibm iseries

LouwMarais:
### Relevant information

Added a new External IBM iSeries entity definition with the main goal being that our custom insights metric generated with an attribute of includeInIseriesEntity should be used to automatically create our IBM iSeries as an Entity and visible on distributed tracing maps

### Checklist

### DIFF
--- a/entity-types/ext-ibm_iseries/definition.yml
+++ b/entity-types/ext-ibm_iseries/definition.yml
@@ -1,0 +1,18 @@
+domain: EXT
+type: IBM_ISERIES
+
+synthesis:
+  name: hostName
+  identifier: hostName
+  encodeIdentifierInGUID : true
+  conditions:
+  - attribute: includeInIseriesEntity
+    present: true
+  tags:
+    systemName:
+      multiValue: false
+    ipAddress:
+      multiValue: false
+configuration:
+  alertable: true
+  entityExpirationTime: DAILY


entityBot (approved):

entityBot (approved):

entity-types/ext-ibm_iseries/definition.yml
```diff
@@ -0,0 +1,18 @@
+domain: EXT
+type: IBM_ISERIES
+
+synthesis:
+  name: hostName
+  identifier: hostName
```
otaviocarvalho, line 6: This rule is too broad and will match any datapoints for all customers of the company that include a `hostName`. Could you please share the use case and metric with us on the `#help-relationship-platform` internal channel?

LouwMarais: Thanks for the feedback. I hoped that the condition that I add an attribute of "includeInIseriesEntity" with any value in a log/custom metric would be the only time that the entity would be used. This is an example of how we would send an event to tag to this entity

`
curl --location 'https://insights-collector.newrelic.com/v1/accounts/<accountIdHere>/events' \
--header 'Content-Type: application/json' \
--header 'X-Insert-Key: <customInsgihtKeyHere> \
--data '{
    "eventType": "MyCustomInsightEvent",
    "hostName": "MyUniqueHostname",
    "includeInIseriesEntity" : "*YES",
    "systemName": "MyUniqueSystemName",
    "ipAddress": "10.11.12.13",
    "otherAttribute1": "otherValue1",
    "otherAttribute2": "otherValue2"
}'
`

LouwMarais: I assume '#help-relationship-platform' is a slack channel, if you can also let me know how to get invited to that happy to chat more.

entityBot (approved):

entityBot (approved):

entity-types/ext-ibm_iseries/definition.yml
LouwMarais: @otaviocarvalho let me know how I can get in touch with you if need to discuss this one more or for further advice. I am not sure how to remove the label after providing the feedback. 
Thanks for your time!

otaviocarvalho: hi @LouwMarais, my apologies, we usually work w/ New Relic field engineers and I didn't realize you were an external client.

You reasoning is indeed correct, adding `includeInIseriesEntity` should be enough to differentiate it from other use cases. If there are other IBM entities already created, I would encourage you to attempt to keep the telemetry aligned among them. e.g. 

The amount of information we include into the rule is really important to us for two reasons:
1. To make sure the new rule has a low likelihood of changing in the future (and breaking existing implementations based on it)
2. To make sure the underlining queries we issue to our internal database systems are not too broad (retrieving too many results/entries per query and impacting our internal systems)

entityBot (approved):

entityBot (approved):

vlopeziglesias (approved):

PR #1959, opened 2025-03-11
NR-347971 |  OpenLineage event consumer writer 

devendra-nr:
Relevant information

Added domain and entities for [Data Pipeline Observability](https://newrelic.atlassian.net/wiki/spaces/DATA/pages/3833725531/Data+Pipeline+Observability+1.0+-+Change+Design+Doc) product

Checklist

 I've read the guidelines and understand the acceptance criteria.
 The value of the attribute marked as identifier will be unique and valid.
 I've confirmed that my entity type wasn't already defined. If it is I'm providing an explanation above.

### DIFF
--- a/entity-types/dpo-olin_dataset_versions/definition.yml
+++ b/entity-types/dpo-olin_dataset_versions/definition.yml
@@ -35,6 +35,9 @@ synthesis:
       dataset.updatedAt:
         entityTagName: olin.dataset.updatedAt
         multiValue: false
+      hash:
+        entityTagName: olin.dataset.hash
+        multiValue: false
 configuration:
-  entityExpirationTime: MANUAL
+  entityExpirationTime: QUARTERLY
   alertable: true
--- a/entity-types/dpo-olin_datasets/definition.yml
+++ b/entity-types/dpo-olin_datasets/definition.yml
@@ -32,5 +32,5 @@ synthesis:
         entityTagName: olin.dataset.name
         multiValue: false
 configuration:
-  entityExpirationTime: MANUAL
+  entityExpirationTime: QUARTERLY
   alertable: true
--- a/entity-types/dpo-olin_job_versions/definition.yml
+++ b/entity-types/dpo-olin_job_versions/definition.yml
@@ -38,6 +38,9 @@ synthesis:
       pipeline.name:
         entityTagName: olin.pipeline.name
         multiValue: false
+      hash:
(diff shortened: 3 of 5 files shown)


entityBot (approved):

entityBot (approved):

entityBot (approved):

entityBot (approved):

entity-types/dpo-olin_dataset_versions/definition.yml
```diff
@@ -0,0 +1,40 @@
+domain: DPO
+type: OLIN_DATASET_VERSIONS
+synthesis:
+  rules:
+  - compositeIdentifier:
+        separator: ":"
+        attributes:
+          - dataset.namespace
+          - dataset.name
+          - updatedAt
```
naxhh, line 10: why updatedAt?

Wouldn't this create a new entity every time the field is updated? is that desired?
Can we talk about how many entities are we expecting here?

entity-types/dpo-olin_dataset_versions/definition.yml
```diff
@@ -0,0 +1,40 @@
...
+        entityTagName: olin.dataset.namespace
+        multiValue: false
+      dataset.name:
+        entityTagName: olin.dataset.name
+        multiValue: false
+      dataset.updatedAt:
+        entityTagName: olin.dataset.updatedAt
+        multiValue: false
+configuration:
+  entityExpirationTime: MANUAL
```
naxhh, line 39: Entities that use synthesis can't be manual they must have a TTL

naxhh (changes requested):
Mostly concerned about the updatedAt as part of the identifier.

And that we are saying these are entities without TTL but using synthesis.

naxhh:
marked as internal review see https://newrelic.slack.com/archives/C07ELLRFDG9/p1741961591913679?thread_ts=1741942875.625519&cid=C07ELLRFDG9

entity-types/dpo-olin_dataset_versions/definition.yml
devendra-nr: For our use case we want to store the history of changes to dataset schema for very long period 12+ months.
Also the number of entities generated are quite low ~10 per month per account.
So we want to handle the delete logic manually in our codebase.

entityBot (approved):

entityBot (approved):

PR #1999, opened 2025-04-10
Fix issues in existing definitions

vlopeziglesias:
### Relevant information

Some of the definitions in the repo contained issues that were not reported at the moment of merging the PR. Fixing them. `hashAlgorithm` must always be provided and `attribute` must be a string representing the tagKey that will be used for the candidate field

### Checklist

### DIFF
--- a/relationships/synthesis/EXT-SERVICE-to-MEMCACHED.yml
+++ b/relationships/synthesis/EXT-SERVICE-to-MEMCACHED.yml
@@ -66,8 +66,8 @@ relationships:
           fields:
             - field: endpoint
               attribute: net.peer.name
-            - field: port
-              attribute: 11211 # Default port for memcached
+#            - field: port
+#              attribute: 11211 # Default port for memcached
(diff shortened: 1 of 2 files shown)


entityBot (approved):

relationships/synthesis/EXT-SERVICE-to-MEMCACHED.yml
```diff
@@ -66,8 +66,8 @@ relationships:
           fields:
             - field: endpoint
               attribute: net.peer.name
-            - field: port
-              attribute: 11211 # Default port for memcached
+#            - field: port
+#              attribute: 11211 # Default port for memcached
```
naxhh, line 70: imho this and removing the rule is the same.

Now we will send a candidate that doesn't match anything never because it requires endpoint and port to match: https://github.com/newrelic/entity-definitions/blob/main/relationships/candidates/MEMCACHED.yml

PR #2037, opened 2025-05-21
feat: store contains kiosk relationship

Kav91:
Intended for STAGING only at this point for testing.
As per internal discussions around point of sales systems cc

Both `STORE` and `KIOSK` contain a `store.number` attribute in their data as well as set as a tag.
So the intention is that these become related (STORE containing KIOSK) based on this field.

Please let me know if the relationship looks sound based on the above or not.

### Checklist


entityBot (approved):

entityBot (approved):

entityBot (approved):

nr-cmorenoin (approved):

naxhh (approved):
Some comments overall LGTM. If you can share with us privately some telemetry and entities we can make sure it will work as you expect but it does look correct.

PR #2094, opened 2025-07-03
Nordstorm changes to test in staging env.

sbankuru-nr:
### Checklist


sbankuru-nr:
Tested in production in latest release.

PR #2273, opened 2025-09-28
fix: Make keda identifiers more specific

hannahsohkim:
### Relevant information
Make keda identifiers more specific so we drop webhook entities

### Checklist


entityBot (approved):

entityBot (approved):

entity-types/ext-keda/definition.yml
```diff
@@ -13,6 +13,20 @@ synthesis:
...
+        attributes:
+          - cluster_name
+          - pod
+      encodeIdentifierInGUID: true
+      name: pod
+      conditions:
+        - attribute: metricName
+          prefix: keda_
+        - attribute: k8s.pod.label.app.kubernetes.io/name
+          value: keda-operator
```
dbudziwojskiNR, line 29: I don't think we need this definition since the `k8s.pod.label.app.kubernetes.io/name` label would only appear in an OTEL cluster which would have a different composite ID.

entityBot (approved):

entityBot (approved):

PR #2317, opened 2025-10-06
Added ext-store, updated relationships, dashboards and metrics

TeonLucas:
### Relevant information

Adjusting entities for RMP project

### Checklist

* These entities are already in production.  Correcting relationships, tags, filter by, metrics, etc.


entity-types/ext-sesame/definition.yml
```diff
@@ -6,41 +6,25 @@ synthesis:
   conditions:
```
nr-cmorenoin, line 6: I have the feeling this conditions are very wide and could match false positives.

Is there a way you can make it more explicit? Are all the XMLRPC events with node=true EXT-SESAME?

entity-types/ext-store/definition.yml
```diff
@@ -0,0 +1,30 @@
+domain: EXT
+type: STORE
+synthesis:
+  name: store.number
+  identifier: store.number 
+  conditions:
```
nr-cmorenoin, line 6: Same here. Is there a posibility to make the conditions more explicit to narrow the scope of this entity type?

nr-cmorenoin (commented):
Is it possible to split this PR on different PRs so we can roll the changes out gradually?

TeonLucas:
@nr-cmorenoin I have made all the RMP entity definitions more distinct, requiring store.number and WayStation.
As long as we are comfortable now that it won't synthesize these for anyone else, it is still a WIP so easiest to iterate once we see these updates on the RMP production data.

PR #10, opened 2020-12-11
Improve documentation for tags and example definition

IreneMLC:
### Relevant information

- Added tags and golden tags information on the README. 
- Added comments to the example definition fields as requested. 

### Checklist

* [x] I've read the guidelines and understand the acceptance criteria.
* [x] The value of the attribute marked as `identifier` will be unique and valid. 
* [x] This is not my first contribution, or I've reached out to the team by opening an issue before
 opening this PR. 
* [x] I've confirmed that my entity type wasn't already defined. If it is I'm providing an
 explanation above.

PR #2362, opened 2025-10-22
removing preview rules

Philip-R-Beckwith:
### Relevant information
Currently customers are seeing duplicated entities being created because a single host entity can match multiple rules. 

The issue is that currently System metrics are being caught by a rule that uses the K8s.node.name as the entity identifier, while process metrics are being synthesized by a rule that is using host.id. If the rule using K8s.node.name is removed, there is another rule lower in the chain that would synthesize system metrics into entities using host.id which would point to the same entity as the process metrics. 

To prevent the entity duplication issue we either have to remove the rule using K8s.Node.name for system metrics or add a rule to use K8s.node.name for process metrics at the top of the chain. 

Since the rule with K8s.node.name was intended to be a temporary workaround for a bug that no longer exists. We've come to the conclusion that removing this rule is the best course of action. 

### Checklist


entityBot (changes requested):
beep boop bop.

 ```
* Error VALIDATING domainType: 'INFRA-HOST' on synthesis - Exception: No entity matches 'INFRA-HOST' synthesis rules for the given test data: 'MetricRaw.json file - #9' data point. Please review test data and synthesis rules.

* Error VALIDATING domainType: 'INFRA-HOST' on synthesis - Exception: No entity matches 'INFRA-HOST' synthesis rules for the given test data: 'MetricRaw.json file - #10' data point. Please review test data and synthesis rules.
```

entityBot (approved):

entityBot (approved):

dbudziwojskiNR (dismissed):

entityBot (dismissed):

Philip-R-Beckwith:
@bgalmesnr Added ownership.

entityBot (approved):

vlopeziglesias (approved):

PR #2368, opened 2025-10-23
fix no_op action on oracle infra db

naxhh:
### Relevant information

Fixes https://github.com/newrelic/entity-definitions/pull/2366

this should have been triggered on the checks but wasn't... but is triggering in new pr's it seems

### Checklist


entityBot (approved):

naxhh:
PR fails because runs with master. merging to have it there.

PR #2438, opened 2025-11-12
Fix fragments

abhishuraina:
### Checklist


relationships/synthesis/INFRA-ELASTICSEARCHCLUSTER-to-ELASTICSEARCHNODES.stg.yml
```diff
@@ -68,4 +67,4 @@ relationships:
         extractGuid:
           attribute: entityGuid
           entityType:
-            attribute: nr.entityType
+            attribute: ELASTICSEARCHNODE
```
davidgonzalez-nr: as it is no longer an attribute but a value, it should be `value: ELASTICSEARCHNODE` instead. Check out example in our [docs](https://github.com/newrelic/entity-definitions/blob/5bf78fc6865dec9c424174aeae91c27a404e9f00/docs/relationships/relationship_synthesis.md#infra-na-for-extractguid)

davidgonzalez-nr (approved):

PR #2458, opened 2025-11-17
Collapse case-insensitive anyOf

jeffalder:
### Relevant information

The `anyOf` condition [defaults to case-insensitivity](https://github.com/newrelic/entity-definitions/blob/main/docs/relationships/relationship_synthesis.md#operations). That means that a condition that defines multiple options that differ only on case will waste effort. In several cases, this is the difference between a single-item lookup and a stream operation. 

These changes should make relationship synthesis more efficient.

### Checklist

* n/a The value of the attribute marked as `identifier` will be unique and valid. 
* n/a I've confirmed that my entity type wasn't already defined. If it is I'm providing an explanation above.


jeffalder:
@nr-cmorenoin @naxhh I've redone this on latest main.

PR #2484, opened 2025-11-25
Validate relationship synthesis rule expires field

vlopeziglesias:
### Checklist

### DIFF
--- a/relationships/synthesis/APM-APPLICATION-to-DATABASE.yml
+++ b/relationships/synthesis/APM-APPLICATION-to-DATABASE.yml
@@ -8,7 +8,7 @@ relationships:
       - attribute: metricName
         regex: "^datastore/instance/(?:mysql|postgres|mssql|oracle|mariadb|jdbc)/[^/;]*/.*"
     relationship:
-      expires: P1YT75M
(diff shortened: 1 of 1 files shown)


package-lock.json
vlopeziglesias: we don't need `package.json` at this path, also it is empty

relationships/synthesis/APM-APPLICATION-to-DATABASE.yml
```diff
@@ -8,7 +8,7 @@ relationships:
       - attribute: metricName
         regex: "^datastore/instance/(?:mysql|postgres|mssql|oracle|mariadb|jdbc)/[^/;]*/.*"
     relationship:
-      expires: PT75M
+      expires: PT72H
```
naxhh: why we go here from 75 minutes to 72 h? what is the context i'm missing?

vlopeziglesias: I don't recall changing this one, I will revert it, maybe it was the one I was testing with locally

entityBot (dismissed):

naxhh (approved):

PR #2507, opened 2025-12-04
Stg us unique validation split

naxhh:
### Relevant information

**NOTE** Test will fail because is using the scripts on main. Run locally it passes all checks

Change the way we validate uniquness of ruleName

Now we check STG files on their own.  This is not perfect as a STG file may override a prod one but since this is mostly for metrics should be fine.

This way we allow the stg file to use the same ruleName as the file it overrides which should be more clear.

### Checklist

### DIFF
--- a/validator/tools/validate_rules.js
+++ b/validator/tools/validate_rules.js
@@ -78,9 +78,9 @@ let PROD_SYNTHESIS_RULES;
 let STG_SYNTHESIS_RULES;
 
 function validateRuleName(ruleName, entityType, isProd) {
-  var lookup = PROD_SYNTHESIS_RULES
+  var lookup = STG_SYNTHESIS_RULES
   if (isProd) {
-    var lookup = STG_SYNTHESIS_RULES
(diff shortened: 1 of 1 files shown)


entityBot (approved):

validator/tools/validate_rules.js
```diff
@@ -74,20 +74,26 @@ async function checkDashboardExists (domain, type, dashboardFile) {
...
 
-function validateRuleName(ruleName, entityType) {
-  if (ALL_SYNTHESIS_RULES.has(ruleName)) {
-    const existing = ALL_SYNTHESIS_RULES.get(ruleName);
-    throw new Error(`Duplicate ruleName '${ruleName}' - already used in ${existing}`);
+function validateRuleName(ruleName, entityType, isProd) {
+  var lookup = PROD_SYNTHESIS_RULES
+  if (isProd) {
+    var lookup = STG_SYNTHESIS_RULES
   }
```
vlopeziglesias: the other way around, right?

PR #2510, opened 2025-12-05
Updated the NRDOT Host Receiever Origin.

pkudikyala:
**Description:** Updated the NRDOT Host Receiver origin from OnHost Integration to OpenTelemetry.

### Checklist


nr-cmorenoin (changes requested):
It looks good to me but it looks like a good moment to add the owner of the entity

https://pages.datanerd.us/site-engineering/nr-platform-docs/entity-platform/new-entity-type/reviewing-entity-definitions.html#how-ownership-works

PR #2536, opened 2025-12-15
chore: Add TTL to tags.* prefix for K8s OpenTelemetry entities to prevent duplicate tag accumulation

NRhzhao:
### Relevant information

When customers use the OpenTelemetry collector to add custom tags via the tags.* prefix (e.g., tags.version, tags.environment), these tags are indexed on entities with the prefix stripped (e.g., version, environment).

Currently, when tag values change (e.g., version: "10" → version: "11"), both the old and new values persist indefinitely on the entity, creating confusing duplicate tags that accumulate over time.

### References
From the official documentation (`docs/entities/synthesis.md:391`):
The prefix `tags.` will always index tags and doesn't need to be specified as part of this structure. However, it's possible to override its default configuration by including the prefix here.

### Solution

Added ttl: P1D (1-day expiration) to the prefixedTags section for the tags. prefix across all OpenTelemetry synthesis rules. This allows old tag values to expire automatically after 24 hours of not being reported, while current values remain.

Implementation Details

- Scope: All OpenTelemetry metrics synthesis rules (api.metrics.otlp and OtlpInfrastructureEvent)
- Pattern applied:
```yaml
prefixedTags:
  tags.:
    ttl: P1D
```
- Entity types modified: 14 entity types (K8s cluster, deployments, pods, containers, hosts, CoreDNS, etc.)
- Not modified: Log-based rules (api.logs.otlp) and non-OpenTelemetry rules (nri-bundle integrations)

Before this change:
1. Customer deploys OTel collector with tags.version: "10"
2. Entity shows tag: version: 10
3. Customer upgrades collector to tags.version: "11"
4. Entity shows both tags indefinitely: version: 10, version: 11 

After this change:
1. Customer deploys OTel collector with tags.version: "10"
2. Entity shows tag: version: 10
3. Customer upgrades collector to tags.version: "11"
4. Entity initially shows both: version: 10, version: 11
5. After 24 hours, old value expires
6. Entity shows only: version: 11 

Backward Compatibility
  - Only behavior change: old tag values expire instead of persisting foreverV

### Checklist

### DIFF
--- a/entity-types/ext-coredns/definition.yml
+++ b/entity-types/ext-coredns/definition.yml
@@ -80,4 +80,4 @@ configuration:
 
 ownership:
   primaryOwner:
-    teamName: "no owner"
+    teamName: "K8s Agents"
--- a/entity-types/infra-kubernetescluster/definition.yml
+++ b/entity-types/infra-kubernetescluster/definition.yml
@@ -113,4 +113,4 @@ synthesis:
 
 ownership:
   primaryOwner:
-    teamName: "no owner"
+    teamName: "K8s Agents"


entityBot (approved):

dbudziwojskiNR (dismissed):

NRhzhao:
Please merge this as soon as passed review, thank you!

entityBot (approved):

dbudziwojskiNR (approved):

PR #2615, opened 2026-02-03
fix:  Added new rules for custom events in Oracle & SQL Server entity definitions

spathlavath:
### Checklist


spathlavath:
@naxhh i'm unable to sign it. i've signed it earlier but not sure why it's required again? can you please help me?

PR #2616, opened 2026-02-03
fix existing k8s relationship issues 

maxa101:
### Relevant information

Entity 
Fix golden metrics for **replicaSet** and **Namespace** 

Relationships
Update Relationships for **Namespace** to use candidates 
Improve **Pod** and **Host** Relationship
Fix **ReplicaSet** relationships and synthesis rules to avoid duplicates 

### Checklist


TmNguyen12 (dismissed):

relationships/synthesis/INFRA-HOST-to-INFRA-KUBERNETES_POD.stg.yml
```diff
@@ -10,10 +10,13 @@ relationships:
...
-          attribute: host.guid
-          entityType:
-            value: HOST
+        lookupGuid:
+          candidateCategory: HOST
+          fields:
+            - field: host.name
+              attribute: nodeName
+            - field: k8s.clusterName
+              attribute: clusterName
```
vlopeziglesias: There is no `HOST` candidate definition with these fields unless I am missing something. So this will not work, could you check again?

kondracek-nr (approved):

rajrohanyadav (approved):
Some changes are in staging files vs some changes are in prod files. I'm not sure but just wanted to point out if that is what you wanted. 
Other than that LGTM.

PR #2900, opened 2026-06-03
fix: remove k8s.pod.name exclusion from Host→HAProxy relationship rule

gmanandhar-nr:
Remove the k8s.pod.name: present: false condition to match the production NGINX Host relationship pattern (INFRA-HOST-to-INFRA-NGINXSERVER.yml) which works without k8s exclusion. In K8s scenarios, both Host and Pod relationships can coexist (same pattern as NGINX).

Only a small change, and change is only in staging. won't impact production.


entityBot (changes requested):
beep boop bop.

gmanandhar-nr:
It took some time to reflect in staging, but we were able to see the host to haproxy ("hosts") relationship without deploying this change

PR #2914, opened 2026-06-08
feat: Add tag for OTel Java agent for EXT service

mdaspute-nr:
### Relevant information
  Added two synthesis rules in definition.stg.yml to detect whether an EXT_SERVICE entity is using OTel Java agent 1.x or 2.x.

  Rules:
  - ext_service_otel_jvm_agent_1x: fires on process.runtime.jvm.* metrics
  - ext_service_otel_jvm_agent_2x: fires on jvm.* metrics

  Both copy `otel.library.version` into tag `jvmInstrumentationVersion`
  with multiValue: false and ttl: P1D.

#### Api Review Board (ARB)

Any pull request with changes to this repository has to be linked with an ARB ticket, which has to be approved before merging.

If you don't know about the ARB process check [this](https://newrelic.enterprise.slack.com/docs/T02D34WJD/F08AW240NSV)

ARB Jira ticket:

### Checklist

### DIFF
--- a/entity-types/ext-service/definition.stg.yml
+++ b/entity-types/ext-service/definition.stg.yml
@@ -386,6 +386,42 @@ synthesis:
       service.namespace:
 
 
+    # Detects OTel Java agent 1.x — fires when process.runtime.jvm.* metrics arrive
+  - ruleName: ext_service_otel_jvm_agent_1x
+    identifier: service.name
+    name: service.name
+    encodeIdentifierInGUID: true
+    conditions:
+    - attribute: service.name
+      present: true
+    - attribute: instrumentation.provider
+      value: opentelemetry
+    - attribute: metricName
+      prefix: process.runtime.jvm
+    tags:
+      otel.library.version:
+        entityTagName: jvmInstrumentationVersion
+        multiValue: false
+        ttl: P1D
+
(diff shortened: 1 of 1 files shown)


entityBot (changes requested):
beep boop bop.

entity-types/ext-service/definition.stg.yml
```diff
@@ -386,6 +386,42 @@ synthesis:
       service.namespace:
 
 
+    # Detects OTel Java agent 1.x — fires when process.runtime.jvm.* metrics arrive
+  - ruleName: ext_service_otel_jvm_agent_1x
+    identifier: service.name
```
alanwest, line 391: I do not think these rules will match any data points. I believe the first matching rule in the file takes precedence. Any JVM metrics will match one of the earlier rules. These rules would need to appear earlier in the file.

I'd also want to know from the entity platform team whether there are any performance concerns with this approach. If we adopt this pattern then more teams may want to create many more rules matching different metrics and conventions. In addition to performance concerns this could become very difficult to maintain. Every single rule needs to be kept in sync with one another. All the tags added by the "standard" OTel rule would need to be duplicated by this new rule.

PR #2942, opened 2026-06-16
fix(validator): relationship-synthesis rule-name uniqueness per env

nr-ksteinbach:
The relationship-synthesis validator (validator/tools/validate_relationship_synthesis_rules.js) gathers every `.yml` file under `relationships/synthesis/` into a single flat list and checks `name` uniqueness against one global Map. That doesn't match the runtime semantics documented in README.md:

> `.stg` file will **replace** [the] original in STG while original will be used for
> PRODUCTION

So `INFRA_X-to-Y.yml` and `INFRA_X-to-Y.stg.yml` never co-exist in any deployed environment, yet the validator treats them as if they did. This forces authors to suffix override identifiers (e.g. `mySuperRule -> mySuperRuleStgOverride`) purely to satisfy CI, even when the override's intent is to *be* the same rule with one specific change for staging — and creates a needless rename on every promotion.

This fix introduces a new helper, `getRelationshipSynthesisDefinitionsByEnvironment`, that partitions definition files into the two file sets that load in each environment (prod = non-.stg.yml; staging = .stg.yml override if present otherwise the .yml). The validator now runs the existing per-rule checks once per environment with its own name-uniqueness map. Cross-environment collisions are no longer flagged; in-environment collisions still fail (now with the env name in the error message for clarity).

### DIFF
--- a/validator/tools/utils.js
+++ b/validator/tools/utils.js
@@ -30,6 +30,62 @@ module.exports = {
       }
     }));
   },
+
+  // Returns relationship synthesis definitions partitioned into the file sets
+  // that load in each environment, mirroring the override mechanic described
+  // in README.md ("`.stg` file will replace original in STG while original
+  // will be used for PRODUCTION"):
+  //   - prod:    every non-.stg.yml file
+  //   - staging: for each base name, the .stg.yml override if present,
+  //              otherwise the .yml; plus any .stg.yml that has no .yml peer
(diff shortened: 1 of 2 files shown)


entityBot (changes requested):
beep boop bop.

entityBot (approved):

bgalmesnr (approved):

bgalmesnr (approved):

PR #2944, opened 2026-06-16
fix(postgresqlinstance): OTel fallback using service.instance.id

nr-ksteinbach:
The existing postgresqlreceiver rule keys synthesis off `postgresql.endpoint`, which the contrib receiver has never emitted as a resource attribute (verified across v0.85 → v0.153). With default contrib output, no POSTGRESQLINSTANCE entity is currently created.

Add a fallback rule that keys off `service.instance.id` (added to the receiver in v0.153.0, in `host:port` form — the exact value the existing rule expects), guarded by `postgresql.endpoint present: false` so it stays mutually exclusive with the existing rule. If a future receiver version starts emitting `postgresql.endpoint`, the existing rule takes precedence and we don't double-mint entities with different GUIDs.

### DIFF
--- a/entity-types/infra-postgresqlinstance/definition.stg.yml
+++ b/entity-types/infra-postgresqlinstance/definition.stg.yml
@@ -22,6 +22,24 @@ synthesis:
         otel.library.name:
           entityTagName: instrumentation.name
 
+    # Fallback for OTel postgres metrics that lack `postgresql.endpoint`.
+    - ruleName: infra_postgresqlinstance_service_instance_id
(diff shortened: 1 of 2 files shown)


entityBot (changes requested):
beep boop bop.

entityBot (dismissed):

entityBot (approved):

entityBot (approved):

spathlavath (approved):
LGTM

bgalmesnr (approved):

PR #2967, opened 2026-06-23
revert(stg): remove EXT-SERVICE container lookup variant (PR #2940)

nr-ksteinbach:
Reverts #2940 (mine) - it was working fine for K8s Containers (nice), but sadly is causing a negative impact on docker / ECS containers -> need to back out until a proper fix is prepared

Re NR-576629.

PR #2940, opened 2026-06-13
feat(stg): add candidate-based EXT-SERVICE container lookup variant

nr-ksteinbach:
See Bug https://new-relic.atlassian.net/browse/NR-576629 and follow the linked slack conversation.

This PR add INFRA_DOCKER_CONTAINER-to-EXT_SERVICE.stg.yml mirroring the prod file:
- extInfraContainerEXTServiceByContainerIdLookup (with a **lookupGuid** against the CONTAINER candidate keyed on container.id, so the source GUID resolves to the existing K8s-integration CONTAINER entity instead of dangling on a fabricated FARM_HASH(container.id))
- nrInfraDockerContainerEXTService (unchanged copy of the legacy NR infra-agent **buildGuid** path)

PR #2977, opened 2026-06-25
feat(stg): relate K8s container to EXT-SERVICE by container.id (NR-576629)

nr-ksteinbach:
**Draft for review** — the proper fix for NR-576629, replacing the approach reverted in #2967 

> ### ⚠️ Reviewer note — review the prod-vs-stg diff, not the raw file count
> GitHub shows this PR as **2 fresh files, +134 lines**. But `INFRA_KUBERNETES_CONTAINER-to-EXT_SERVICE.yml` already exists in prod, and the new `.stg` file is a staging copy of it. The meaningful change is the **diff between the existing prod file and the new `.stg` file** — which is a single added rule (the first 71 lines are identical):
>
> ```diff
> @@ relationships/synthesis/INFRA_KUBERNETES_CONTAINER-to-EXT_SERVICE.yml -> .stg.yml @@
> +  - name: extK8sContainerEXTServiceByContainerId
> +    version: "1"
> +    origins: [ OpenTelemetry ]
> +    conditions:
> +      - { attribute: eventType, anyOf: [ "Log", "Metric", "Span" ] }
> +      - { attribute: entity.type, anyOf: [ "SERVICE" ] }
> +      - { attribute: k8s.cluster.name, present: true }
> +      - { attribute: k8s.pod.name, present: true }
> +      - { attribute: container.id, present: true }
> +      - { attribute: k8s.container.name, present: false }
> +    relationship:
> +      expires: PT75M
> +      relationshipType: HOSTS
> +      source:
> +        lookupGuid:
> +          candidateCategory: K8S_CONTAINER_BY_ID
> +          fields: [ { field: containerId, attribute: container.id } ]
> +      target:
> +        extractGuid: { attribute: entity.guid }
> ```
> The two pre-existing rules are reproduced verbatim so staging behavior is a strict superset of prod. The other new file is the `K8S_CONTAINER_BY_ID` candidate.

### Problem
An OTel K8s service's telemetry carries `container.id` but **not** `k8s.container.name`, so the two existing `buildGuid` rules in `INFRA_KUBERNETES_CONTAINER-to-EXT_SERVICE` (whose source GUID is the `cluster:ns:pod:container.name` composite) never fire — the container→service edge is missing.

### Why not the #2940 approach
#2940 resolved the container via the shared `CONTAINER` candidate, whose `onMiss: CREATE_UNINSTRUMENTED` mints a placeholder on every miss. Since dockerstats/ECS containers don't carry a `*.containerId` tag, every OTel-docker/ECS service got wired to an uninstrumented placeholder (confirmed live in staging). It was reverted in #2967.

### This approach
1. **New candidate `K8S_CONTAINER_BY_ID`** (`onMiss: NO_OP`) — resolves `container.id` → the K8s container's `k8s.containerId` tag. On a miss it **discards** rather than materializing a placeholder.
2. **Additive 3rd rule** in the K8s-owned `INFRA_KUBERNETES_CONTAINER-to-EXT_SERVICE.stg.yml`, gated to K8s telemetry that has `container.id` but lacks `k8s.container.name`, looking up via that candidate.

Worst case is *no edge* (the pre-existing NR-576629 symptom), never a wrong edge or junk entity.

### Notes / open questions for reviewers
- Candidate `category` name (`K8S_CONTAINER_BY_ID`) is a placeholder — happy to rename to your convention.
- Resolution relies on the container carrying `k8s.containerId`, which is set by KSM (`kube_pod_container_info`) and the infra-agent `K8sContainerSample`, but **not** by cadvisor/kubeletstats-only synthesis. Those clusters safely no-op (no edge).
- Ideal longer-term fix is upstream: have the OTel collector add `k8s.container.name` (k8sattributesprocessor) so the existing composite `buildGuid` rule fires with no candidate at all.


entityBot (changes requested):
beep boop bop.

entityBot (approved):

PR #2994, opened 2026-06-30
feat(kafka): add kafka.cluster.id tags and relationship rules for self-hosted Kafka

shashank-reddy-nr:
### Relevant information

Add `kafka.cluster.id` (Kafka cluster UUID) support to entity synthesis tags and relationship rules for self-hosted Kafka. This complements the existing `kafka.cluster.name`-based rules and is required now that `kafkametricsreceiver` emits `kafka.cluster.id` as a resource attribute.

#### Changes

**`entity-types/infra-kafkacluster/definition.stg.yml`**
- Added `kafka.cluster.id:` tag to all 8 OTel synthesis rules (jmxmetrics, jmxscraper, kafkametrics, prometheus — both Kafka and JVM variants, plus logs rule).
- Added `kafka.cluster.id` to `goldenTags` so it appears prominently on the KAFKACLUSTER entity.

**`entity-types/infra-kafkatopic/definition.stg.yml`**
- Added `kafka.cluster.id:` tag to the existing `infra_kafkatopic_otel_kafkametrics_stg` rule so KAFKATOPIC entities created by `kafkametricsreceiver` carry the cluster UUID as a tag (enabling candidate lookup by UUID).

**`relationships/candidates/KAFKATOPIC.stg.yml`**
- New lookup predicate: match `INFRA/KAFKATOPIC` by tags `kafka.cluster.id` (→ field `clusterId`) + `topic` (→ field `topic`), with `onMiss: NO_OP`. This enables `lookupGuid candidateCategory: KAFKATOPIC` resolution using the cluster UUID instead of the cluster name.

**`relationships/synthesis/APM-APPLICATION-to-KAFKATOPIC.stg.yml`**
- `apmProducesNRKafkaTopicByClusterIdStg`: NR APM metric `MessageBroker/Kafka/Cluster/{cluster_id}/Topic/{topic}/Produce` → PRODUCES → KAFKATOPIC via `lookupGuid` (clusterId from `metricName__4`, topic from `metricName__6`).
- `apmConsumesNRKafkaTopicByClusterIdStg`: same pattern for `Consume`.

**`relationships/synthesis/EXT-SERVICE-to-KAFKATOPIC.stg.yml`**
- `extServiceProducesSelfHostKafkaTopicClusterIdStg`: OTel producer/client span with `kafka.cluster.id` + `messaging.destination.name` → PRODUCES → KAFKATOPIC via `lookupGuid`.
[...]

PR #3023, opened 2026-07-03
fix(redis-otel): Use NO_OP instead of CREATE_UNINSTRUMENTED for clust…

vagrawal-newrelic:
## Summary

Fixes CI failure introduced by the Redis OTel cluster candidate relationship file. Changes `onMiss` action from `CREATE_UNINSTRUMENTED` to `NO_OP` for the `REDISCLUSTER` candidate category.

## Problem

The `REDISCLUSTER.stg.yml` candidate relationship used `CREATE_UNINSTRUMENTED` which requires a corresponding `uninstrumented-rediscluster/definition.stg.yml` file to exist. This file was missing, causing the `validate_uninstrumented_definitions_exists` CI check to fail for all PRs.

## Fix

Changed to `action: NO_OP` — when an APM service references a Redis cluster that doesn't have an instrumented entity, do nothing instead of creating a placeholder. This matches the approach used by Kafka topics (PR #2994).

## Why NO_OP over CREATE_UNINSTRUMENTED

- Avoids creating potentially confusing placeholder entities for customers
- Simpler — no additional uninstrumented entity definition file needed
- Consistent with the Kafka OTel pattern (`KAFKATOPIC.stg.yml`)
- Relationships still work correctly when both entities are instrumented

### DIFF
--- a/relationships/candidates/REDISCLUSTER.stg.yml
+++ b/relationships/candidates/REDISCLUSTER.stg.yml
@@ -11,6 +11,4 @@ lookups:
     onMatch:
       onMultipleMatches: RELATE_ALL
     onMiss:
-      action: CREATE_UNINSTRUMENTED
-      uninstrumented:
-        type: REDISCLUSTER
+      action: NO_OP


entityBot (approved):

mborroni (approved):

PR #3034, opened 2026-07-09
feat(mysqlnode-otel): Complete OpenTelemetry mysqlreceiver support in Staging

abhinav1602:
Completes **OpenTelemetry Collector [`mysqlreceiver`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/mysqlreceiver)** support for the existing `INFRA-MYSQLNODE` entity. All changes are **staging-only** (`.stg` overrides) — production behavior is unchanged until a later promotion.

**What changed**

- **`entity-types/infra-mysqlnode/definition.stg.yml`** *(new — full staging override)*
  - OTel metric rule: added `server.address` / `server.port` tags (used by the new relationship lookup).
  - Two new **Log synthesis rules** for query-performance events (`event.name` prefix `db.server.`) — one for OTel-contrib `mysqlreceiver`, one for the `newrelic-forks` `nrmysqlreceiver`. The `otel.library.name` condition disambiguates from other DB receivers that also emit `db.server.*`.
  - Metric + logs share the `mysql.instance.endpoint` identifier, so an instance is one entity regardless of signal.
- **`golden_metrics.stg.yml`** — fixed `bufferPoolDataPages` facet (`state` → `kind`); added OTel queries for connected/running threads, max used connections, buffer-pool pages; added OTel-only `bufferPoolUsageBytes` and `uptime`. Counter metrics stay `newRelic`-only (cumulative/monotonic can't take value functions cleanly). At the 10-metric schema cap.
- **`opentelemetry_dashboard.stg.json`** — rebuilt single-page dashboard covering the receiver's default metrics plus a Query Performance section from the log events. Widgets on disabled-by-default metrics are labeled "requires enabling".
- **`tests/Log.stg.json`** *(new)* — sample `db.server.top_query` / `db.server.query_sample` payloads matching the new log rules.
- **`relationships/candidates/MYSQLNODE.stg.yml`** *(new)* — indexes MYSQLNODE entities by `server.address` + `server.port` (mirrors `MSSQLDBINSTANCE`).
- **`relationships/synthesis/OTEL-APPLICATION-to-INFRA-MYSQLNODE.stg.yml`** *(new)* — `CALLS` relationship from OTel-instrumented apps to the MySQL node via distributed-tracing spans (covers both `db.system` and `db.system.name = mysql`). The existing native-APM relationship is untouched.

All metric/attribute names verified against the receiver's `metadata.yaml`.

**Validation**

`npm --prefix validator run check` passes (schemas + synthesis rules + relationship rules).

> ⚠️ **To confirm in staging:** the `server.address` / `server.port` tags that the candidate lookup keys on are captured from telemetry, but the receiver's `metadata.yaml` lists only `mysql.instance.endpoint` as a resource attribute. Before promotion, verify these tags actually populate on the entity (via collector processor / real ingest); otherwise the OTEL-APP → MYSQLNODE relationship won't resolve.


entityBot (dismissed):

relationships/synthesis/OTEL-APPLICATION-to-INFRA-MYSQLNODE.stg.yml
```diff
@@ -0,0 +1,60 @@
+relationships:
+  # OpenTelemetry-instrumented application -> MySQL node, from distributed tracing spans.
```
spathlavath: Currently, we are only establishing synthesis relationships from Native APM to OpenTelemetry (OTel) DB receivers. Note that the Query Comments feature is exclusively supported by native APM agents.

For context on how these are mapped, you can refer to the existing synthesis rules and their tracking candidates for [Oracle](https://github.com/newrelic/entity-definitions/blob/main/relationships/synthesis/APM-APPLICATION-to-INFRA-ORACLEDBINSTANCE.yml) and [MSSQL](https://github.com/newrelic/entity-definitions/blob/main/relationships/synthesis/APM-APPLICATION-to-INFRA-MSSQLINSTANCE.yml).

entityBot (approved):

relationships/synthesis/OTEL-APPLICATION-to-INFRA-MYSQLNODE.stg.yml
abhinav1602: Thanks for pointing out @spathlavath , removed the OTEL APM file.

pkulkarni-newrelic (approved):

bgalmesnr (approved):

PR #3052, opened 2026-07-16
feat: Remove relationship for namespace and replicaset

codylawson:
### Relevant information
Removing relationship between k8s namespace and replicaset. Namespace relationships merged previously had caused a large spike in cpu% during synthesis, replicaset being a larger portion of this. Removing for now while we try to get the rest of the namespace relationships stable.

#### Api Review Board (ARB)

ARB Jira ticket:
https://new-relic.atlassian.net/browse/NR-541144

Business justification doc: https://newrelic.atlassian.net/wiki/spaces/KOT/pages/5297407012/Business+Justification+for+K8s+Entities+relationships

### Checklist

* [ ] I've read the guidelines and understand the acceptance criteria.
* [ ] The value of the attribute marked as `identifier` will be unique and valid.
* [ ] I've confirmed that my entity type wasn't already defined. If it is I'm providing an explanation above.
* [ ] I've linked an ARB ticket & received approval from API Review Board in order to make these changes

PR #3054, opened 2026-07-17
Revert "feat: Remove relationship for namespace and replicaset"

bgalmesnr:
Reverts newrelic/entity-definitions#3052


entityBot (approved):

mborroni (approved):

PR #3050, opened 2026-07-15
feat(gcp): GCP Managed Kafka synthesis rules for KAFKACLUSTER and KAFKATOPIC

abhavanam-blip:
## What

GCP **Managed Service for Apache Kafka** synthesis rules for the staging definitions of `KAFKACLUSTER` and `KAFKATOPIC` (`definition.stg.yml`)

## Why

Managed Kafka metrics from **gcp-polling-v2** carry `instrumentation.provider = gcp`; no existing OTEL/native rule matches them, and the legacy metadata path mints `NA`-typed GUIDs that never register. These shared infra types now need real GCP synthesis rules.

KAFKACLUSTER:
cluster_id identifier: regression-cn-kafka
Screenshot 2026-07-16 at 11 55 01 AM

KAFKATOPIC:
gcp.topic_id identifier:  regression-cn-topic
gcp.cluster_id identifier: regression-cn-kafka
Screenshot 2026-07-16 at 12 02 34 PM

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3057, opened 2026-07-20
fix(gcp): add legacyFeatures to GCP Managed Kafka KAFKACLUSTER/KAFKATOPIC synthesis rules

abhavanam-blip:
## Problem

Follow-up to #3050.

The GCP Managed Kafka synthesis rules added in #3050 **don't register entities**. `KAFKACLUSTER` and `KAFKATOPIC` are legacy infra types whose GUIDs use a **non-standard `NA` type slot** — and without `legacyFeatures`, the entity platform won't mint/register an entity from the synthesis rule. The metric's `entity.guid` stays a dangling `NA` guid and nothing browsable is created (verified: 0 synthesized GCP `KAFKATOPIC`/`KAFKACLUSTER` entities; every existing one is a metadata-path `NA` guid).

Confirmed with the IDP Team: https://newrelic.slack.com/archives/C0FJ72M28/p1784536268575059?thread_ts=1784519694.135909&cid=C0FJ72M28

### DIFF
--- a/entity-types/infra-kafkacluster/definition.stg.yml
+++ b/entity-types/infra-kafkacluster/definition.stg.yml
@@ -247,6 +247,9 @@ synthesis:
     identifier: cluster_id
     name: cluster_id
     encodeIdentifierInGUID: true
+    legacyFeatures:
+      useNonStandardAttributes: true
+      overrideGuidType: true
     conditions:
       - attribute: eventType
         value: Metric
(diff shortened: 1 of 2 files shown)


entityBot (approved):

cnivyanthsai (approved):

mborroni (approved):

PR #3061, opened 2026-07-20
feat: Add staging ext-service rule for collector auto-tagging

agarvin-nr:
### Relevant information

This PR defines a staging rule for `ext-service` entities to auto-tag entities with `otel_collector` when any metric with an `otelcol_` prefix is detected. The purpose of this is to grant the ability to automatically detect when a collector is present in the entity.

### Checklist


entityBot (approved):

ppant-code (approved):

mborroni (approved):

PR #3074, opened 2026-07-22
feat: Add staging OTel log synthesis rule for POSTGRESQLINSTANCE

mraju1:
#### Api Review Board (ARB)

Any pull request with changes to this repository has to be linked with an ARB ticket, which has to be approved before merging.

If you don't know about the ARB process check [this](https://newrelic.enterprise.slack.com/docs/T02D34WJD/F08AW240NSV)

ARB Jira ticket:

### Checklist

### DIFF
--- a/entity-types/infra-postgresqlinstance/definition.stg.yml
+++ b/entity-types/infra-postgresqlinstance/definition.stg.yml
@@ -1,0 +1,92 @@
+domain: INFRA
+type: POSTGRESQLINSTANCE
+
+synthesis:
+  rules:
+    # opentelemetry (postgresqlreceiver); service.instance.id is host:port, enabled by default
+    - ruleName: infra_postgresqlinstance_service_instance_id
+      identifier: service.instance.id # host:port
+      name: service.instance.id
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: eventType
+          value: Metric
+        - attribute: metricName
+          prefix: postgresql.
+        - attribute: instrumentation.provider
+          value: opentelemetry
+      tags:
+        # Environment resource attributes
+        host.type:
+        cloud.provider:
+        cloud.account.id:
+        cloud.region:
+        otel.library.name:
+          entityTagName: instrumentation.name
+
+    # opentelemetry logs (db.server.top_query / db.server.query_sample);
+    # db.system.name is always present on these log events when enabled.
+    # Using service.instance.id (host:port) to match the metric rule above.
+    - ruleName: infra_postgresqlinstance_otel_logs
+      identifier: service.instance.id # host:port
+      name: service.instance.id
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: eventType
+          value: Log
+        - attribute: instrumentation.provider
+          value: opentelemetry
+        - attribute: db.system.name
+          value: postgresql
+      tags:
+        host.type:
+        cloud.provider:
+        cloud.account.id:
+        cloud.region:
+        otel.library.name:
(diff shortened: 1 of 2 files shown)


entityBot (changes requested):
beep boop bop.

* Error VALIDATING domainType: 'INFRA-POSTGRESQLINSTANCE' on synthesis - Exception: No entity matches 'INFRA-POSTGRESQLINSTANCE' synthesis rules for the given test data: 'Log.json file - #1' data point. Please review test data and synthesis rules.
* Error VALIDATING domainType: 'INFRA-POSTGRESQLINSTANCE' on synthesis - Exception: No entity matches 'INFRA-POSTGRESQLINSTANCE' synthesis rules for the given test data: 'Log.json file - #2' data point. Please review test data and synthesis rules.
* We don't think these changes require an API Review Board (ARB) review, so it is not required to merge this PR. Note that all API changes do require one: if you believe this PR needs an ARB review, please create the ticket and add the `ARB:pending` label. Reach `#api-review-board` in case of doubts.

entity-types/infra-postgresqlinstance/definition.stg.yml
```diff
@@ -0,0 +1,92 @@
...
+          prefix: postgresql.
+        - attribute: instrumentation.provider
+          value: opentelemetry
+      tags:
+        # Environment resource attributes
+        host.type:
+        cloud.provider:
+        cloud.account.id:
+        cloud.region:
+        otel.library.name:
```
spathlavath, line 24: @mraju1 Can we update the tags for both metrics and logs?

- Add `otel.library.version` so we can identify customer collector versions.
- Remove `host.type, cloud.provider, cloud.account.id, and cloud.region` since they aren't present in the PostgreSQL receiver's [metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/postgresqlreceiver/metadata.yaml).

nr-ksteinbach: > Remove host.type, cloud.provider, cloud.account.id, and cloud.region since they aren't present

I think these are useful tags. 
They are coming from the resourcedetectionprocessor and are all very popular.

spathlavath: @nr-ksteinbach Thanks for the clarification!

One thing worth flagging though, `resourcedetectionprocessor` actually detects the collector's own VM/host, not the database's. It reads from the collector's local environment (`EC2/GCP/Azure/system metadata`).

So these tags are accurate only when the collector runs on the same VM as Postgres. If it runs on a separate VM (e.g. a `centralized collector`, or a `managed/remote Postgres instance`), `host.type, cloud.provider, cloud.account.id, and cloud.region` would describe the collector's VM, not the DB's - which could be misleading on `POSTGRESQLINSTANCE`.

Given that's a pretty common setup, I'd lean toward dropping them. Happy to hear your thoughts on this!

mraju1: if a value for the a tag attribute entry is not present, does it ignore tagging it or creates a tag with empty value?

spathlavath: @mraju1 As per the synthesis docs: a tag is only created if the attribute is present on the matching data point. If it's absent, no tag gets written (not an empty-value tag).

So keeping `host.type/cloud.provider/etc.` in the rule is safe when the attribute is missing - the real risk we were discussing is only when the attribute is present but describes the wrong VM (collector's, not the DB's).

Also, can you please fix the errors thrown by entityBot?

PR #3079, opened 2026-07-22
fix: Update candidate tagKeys

mborroni:
Reverts newrelic/entity-definitions#3077


entityBot (approved):

PR #3067, opened 2026-07-21
feat: Add OCINATGATEWAY entity definition (staging)

aprasad-sketch:
Add OCI NAT Gateway entity definition with synthesis rules, golden metrics, and summary metrics. **Staging only** (`.stg.yml` / `Metric.stg.json`).

## Entity Type
- Type: `INFRA-OCINATGATEWAY`
- Provider: OCI (Oracle Cloud Infrastructure)
- Namespace: `oci_nat_gateway`
- OCID prefix: `ocid1.natgateway`

## Golden Metrics
- BytesFromNATgw (`oci.nat.gateway.bytes.from.natgw`)
- BytesToNATgw (`oci.nat.gateway.bytes.to.natgw`)
- ConnectionsEstablished (`oci.nat.gateway.connections.established`)
- DropsToNATgw (`oci.nat.gateway.drops.to.natgw`)

## Metric Source
https://docs.oracle.com/en-us/iaas/Content/Network/Reference/nat-gateway-metrics.htm

PR #3097, opened 2026-07-28
fix: Correct OCINATGATEWAY drops metric to DropsFromNATgw

aprasad-sketch:
Follow-up correction to #3067 (already merged).

Verified against **live OCI Monitoring** (`oci monitoring metric list`) in the metrics compartment: the NAT gateway namespace emits **`DropsFromNATgw`**, not `DropsToNATgw`. The Oracle docs reference page (nat-gateway-metrics.htm) lists `DropsToNATgw`, which is inaccurate — no such metric is published. (`highPortUsageWatermark` from the docs is also not emitted.)

## Change
- Golden metric: `dropsToNatGateway` → `dropsFromNatGateway`, query `sum(\`oci.nat.gateway.drops.from.natgw\`)`
- Summary metric: same rename

Without this, the drops golden metric shows **No data**. Staging only (`.stg.yml`).

### DIFF
--- a/entity-types/infra-ocinatgateway/golden_metrics.stg.yml
+++ b/entity-types/infra-ocinatgateway/golden_metrics.stg.yml
@@ -25,12 +25,12 @@ connectionsEstablished:
       from: Metric
       eventId: entity.guid
       eventName: entity.name
-dropsToNatGateway:
-  title: Drops To NAT Gateway
+dropsFromNatGateway:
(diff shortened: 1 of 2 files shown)


entityBot (approved):

dachepallysaikirannewrelic (approved):

vlopeziglesias (approved):

PR #3110, opened 2026-07-30
fix: OTEL APM to Elasticsearch relationship synthesis rule (staging only)

vagrawal-newrelic:
## Summary

Fix the staging synthesis rule for OTEL APM Service → INFRA ELASTICSEARCHCLUSTER relationship. This change is **staging only** (`.stg.yml`).

## Changes

- **origins**: `OpenTelemetry` → `Distributed Tracing` — the rule processes Span events, not Metrics
- **account**: `attribute: accountId` → `lookup: true` — accountId is not available as a span attribute; lookup derives it from the source entity GUID
- **attribute**: `elasticsearch.cluster.name` → `db.namespace` — aligns with current OTel semantic conventions (v1.43+), where `db.namespace` is the standard attribute for database namespace (cluster name for ES)

## Context

The OTel ES instrumentation fetches the cluster name via `GET /` on first connection and sets `db.namespace` on all subsequent spans. Combined with the auto-populated `db.system = elasticsearch` condition, this uniquely identifies the target Elasticsearch cluster entity.

The synthesis rule uses FARM_HASH on the `db.namespace` value to build the target entity GUID, matching the ELASTICSEARCHCLUSTER entity definition.

## Scope

- **Staging only** — `.stg.yml` file, no production impact
- Works for all OTel-instrumented languages (Java, Python, Node.js, Go, .NET)

### Checklist

### DIFF
--- a/relationships/synthesis/OTEL-SERVICE-to-INFRA-ELASTICSEARCHCLUSTER.stg.yml
+++ b/relationships/synthesis/OTEL-SERVICE-to-INFRA-ELASTICSEARCHCLUSTER.stg.yml
@@ -1,11 +1,11 @@
 relationships:
-  # OTEL APM Application calls Elasticsearch Cluster
-  # Matches on elasticsearch.cluster.name attribute in spans
-  # Agent reads cluster name from ES GET / response and enriches spans
+  # OTEL APM Service calls Elasticsearch Cluster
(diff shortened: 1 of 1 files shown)


entityBot (approved):

entityBot (approved):

shashank-reddy-nr (approved):
LGTM

vlopeziglesias (approved):

PR #3112, opened 2026-07-30
fix: guard ECS EC2 OTel host rules against service.name metrics

alanwest:
These rules were added in #2729. This has caused a regression where telemetry from a service is incorrectly synthesized as a host entity. It is of utmost importance that the `service.name present: false` predicate is applied to all rules in this file to avoid future regressions.

@muruganishekar please take note of this and ensure that other rules you may have recently introduced do not suffer from this same problem.

### DIFF
--- a/entity-types/infra-host/definition.yml
+++ b/entity-types/infra-host/definition.yml
@@ -54,6 +54,9 @@ synthesis:
           present: true
         - attribute: host.id
           present: true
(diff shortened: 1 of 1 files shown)


entityBot (approved):

muruganishekar (approved):

naxhh (approved):

PR #3119, opened 2026-08-03
Replicaset namespace relationship

codylawson:
### Relevant information

Adding replicaset <> namespace relationship file with updated condition to filter more entities.

❓ One question that this may raise is using the `regex` condition matcher with an attribute that is a number. Is this approach valid or can regex only be used with string attributes?

#### Api Review Board (ARB)

ARB Jira ticket:
https://new-relic.atlassian.net/browse/NR-541144

Business justification doc: https://newrelic.atlassian.net/wiki/spaces/KOT/pages/5297407012/Business+Justification+for+K8s+Entities+relationships

### Checklist

* [x] I've linked an ARB ticket & received approval from API Review Board in order to make these changes

### DIFF
--- a/relationships/synthesis/INFRA-KUBERNETES_NAMESPACE-to-INFRA-KUBERNETES_REPLICASET.stg.yml
+++ b/relationships/synthesis/INFRA-KUBERNETES_NAMESPACE-to-INFRA-KUBERNETES_REPLICASET.stg.yml
@@ -1,0 +1,54 @@
+relationships:
+  - name: k8sNamespaceContainsReplicaSet
+    version: "1"
+    origins:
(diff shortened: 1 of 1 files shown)


entityBot (approved):

mborroni:
> ❓ One question that this may raise is using the regex condition matcher with an attribute that is a number. Is this approach valid or can regex only be used with string attributes?

The engine converts everything to strings before any condition evaluation occurs so regex conditions work on numeric attributes.

entityBot (approved):

entityBot (approved):

mborroni (approved):

PR #3120, opened 2026-08-03
feat: update ext-service collector auto-detection rule

kb-newrelic:
### Relevant information

- Update to #3061 to match final design from [CDD](https://newrelic.atlassian.net/wiki/x/AwDRXgE). CDD contains details about the design of the rule including the comment above the rule regarding ordering which is a stopgap discussed with EP while the necessary [FR](https://new-relic.atlassian.net/browse/NR-598253) is considered.
- The general idea is for the new rule to be a specialization of the existing `ext_service_service_name_3` to allow tagging OTel Collectors with an entity tag, so that [this manual step](https://docs.newrelic.com/docs/opentelemetry/collector-observability/collector-observability/#add-tag) is no longer necessary to have the UI treat the entity as a special type of EXT-SERVICE.

#### Api Review Board (ARB)

- Will be filed before this rule gets promoted to prod. Our understanding is that testing this rule in staging should not require prior ARB approval.

### Checklist

* [ ] (N/A - see above) I've linked an ARB ticket & received approval from API Review Board in order to make these changes

### DIFF
--- a/entity-types/ext-service/definition.stg.yml
+++ b/entity-types/ext-service/definition.stg.yml
@@ -45,19 +45,6 @@ synthesis:
         entityTagName: language
       telemetry.sdk.version:
 
-  - ruleName: ext_service_otel_collector
-    # produces same entity guid as the default rule applying to all non-otelcol telemetry
-    identifier: service.name
-    name: service.name
-    encodeIdentifierInGUID: true
-    conditions:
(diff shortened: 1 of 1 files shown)


entityBot (approved):

ppant-code (approved):

mborroni (approved):

PR #817, opened 2022-10-21
Recover host synthesis host entity from SAP data source.

edyu002:
modified:   ../definition.yml
	new file:   SAPHost.json

It was decided to use infra-host entity for host information generated by SAP datasource. However, some late changes on host definition broke the process as it check on a data source attribute which are not available for SAP generated metrics. This change is intend to recover the support of synthesizing host entity for data coming from SAP data source.

### Relevant information

Added rules for synthesize SAP host entity.

### Checklist

* [x ] I've read the guidelines and understand the acceptance criteria.
* [ x] The value of the attribute marked as `identifier` will be unique and valid. 
* [x ] I've confirmed that my entity type wasn't already defined. If it is I'm providing an
 explanation above.

PR #3145, opened 2026-08-07
fix: match new MessageBroker/Kafka/Cluster/{id}/[Produce|Consume]/{topic} format

shashank-reddy-nr:
## Summary

`agent-specs` PR #817 realigned the APM per-cluster Kafka metric name to match the existing `Nodes` metric shape (verb before topic, not after):

- Old: `MessageBroker/Kafka/Cluster/{cluster_id}/Topic/{topic}/Produce`
- New: `MessageBroker/Kafka/Cluster/{cluster_id}/Produce/{topic}`

This updates the `apmProducesNRKafkaTopicByClusterIdStg` / `apmConsumesNRKafkaTopicByClusterIdStg` relationship conditions in `APM-APPLICATION-to-KAFKATOPIC.stg.yml` to match. The `metricName__4` (clusterId) / `metricName__6` (topic) field indices are unchanged — the segment count and positions are identical under the new format, so only the regex and comments needed updating.

## Why now

No agent has shipped the old metric format yet — all 4 language-agent implementations (java, dotnet, python, node) are still open drafts. This closes the window before anything ships against a metric shape this relationship rule can no longer match.

## Related

- Spec change: source.datanerd.us/agents/agent-specs/pull/817

### DIFF
--- a/relationships/synthesis/APM-APPLICATION-to-KAFKATOPIC.stg.yml
+++ b/relationships/synthesis/APM-APPLICATION-to-KAFKATOPIC.stg.yml
@@ -171,14 +171,14 @@ relationships:
               - attribute: clusterName
             hashAlgorithm: FARM_HASH
 
-  # Self-hosted Kafka: NR APM MessageBroker/Kafka/Cluster/{cluster_id}/Topic/{topic}/Produce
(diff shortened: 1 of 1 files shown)


entityBot (approved):

entityBot (approved):

PR #3192, opened 2026-08-21
NO-JIRA: ep specific ARB guideline

davidgonzalez-nr:
### Relevant information
Improves pull request template to give better more-specific ep context on ARB


entityBot (approved):

bgalmesnr (approved):

PR #3202, opened 2026-08-26
feat: Enable Host<>Pod relationship by fixing GUID resolution

codylawson:
### Relevant information

Fixes the existing (staging-only) Host/Node<>Pod `HOSTS` relationship in `relationships/synthesis/INFRA-HOST-to-INFRA-KUBERNETES_POD.stg.yml`. Per [NR-609078](https://new-relic.atlassian.net/browse/NR-609078) #1, a file for this relationship already existed but wasn't producing any relationships.

**Root cause:** both sub-rules used `extractGuid` against `host.guid`, an attribute neither `K8sPodSample` nor the OTel `system.*` metric ever carries. I traced this back through the file's full history (#1134 in 2023 through #2414 in Jan 2026) — this bug has existed since the very first commit and has apparently never produced a single relationship.

**Fix, verified against real telemetry (staging account "K8s Agents Team", which runs both classic and OTel/kube-state-metrics k8s nodes):**

- Added a dedicated HOST candidate (`relationships/candidates/HOST.stg.yml`) matching on hostname + cluster name (ALL-match, `NO_OP` on miss), rather than reusing the generic shared hostname-only candidate. This rule fires per-pod across the whole fleet, so falling back to `CREATE_UNINSTRUMENTED` on every miss (as the shared candidate does) would inflate entity/relationship volume — directly counter to this ticket's goal of staying volume-neutral against the Host<>Container relationships being scoped down in the companion PR.
- `k8sHostHostsPod` (classic Kubernetes Integration, `K8sPodSample`): now `lookupGuid` against that candidate, matching `nodeName`/`clusterName`.
- `otelKsmK8sHostHostsPod` (OTel): switched from matching any `system.*` metric to the purpose-built kube-state-metrics `kube_pod_info` metric (cheaper to match, one row per pod per scrape), matching `k8s.node.name`/`k8s.cluster.name` via the same candidate. Target uses `extractGuid` on `entity.guid`, which `kube_pod_info` carries directly.

---

**Update — resolvers reworked after a volume review. The `lookupGuid` approach above now applies only to the OTel/KSM fallback:**

- `k8sHostHostsPod` now uses `buildGuid(FARM_HASH(reportingAgent))` rather than a candidate lookup. `K8sPodSample.reportingAgent` *is* the reporting agent's own `entityKey`, i.e. the HOST GUID's identity string — verified 400/400 on staging account 1 (276 clusters, 60.7M events/30min), 0 missing, 0 mismatched. Removes ~60.5M candidate lookups per 30 minutes in that account alone. Added a `nodeName` regex guard: 0.065% of rows are `Pending` pods with `nodeName=""` from the KSM scraper, which `present: true` does not screen.
- New `otelK8sHostHostsPod` uses kubeletstats `k8s.pod.memory.working_set`, which carries the pod's own node `host.id` — both GUIDs built, no lookup. Covers 7 of the 8 staging OTel clusters that have a HOST entity at all.
- `otelKsmK8sHostHostsPod` (`kube_pod_info` + `lookupGuid`) is **kept** as the fallback for KSM-only collectors, so `HOST.stg.yml` stays as-is. This also refines the "Known gap" below: a KSM-only setup *can* have HOST entities where hostmetrics runs without kubeletstats — `otel-linkerd-demo` (acct 1015429) is a live example, and it is exactly what this fallback serves.

That FARM_HASH reproduces the INFRA GUID identifier is verified against live data, not assumed (the production `hostHostsOtelInfraDockerContainer` rule's edges land on the real HOST entity built from `FARM_HASH(host.id)`). Two caveats for review: the `regex: ".+"` empty-string guard is inferred from the docs rather than observed, and all validation is staging-only.

**Note on a related, previously-closed attempt:** an earlier unmerged PR (#2683) took a similar `lookupGuid`-based approach. I compared against it and pulled in what verified out (the `kube_pod_info` metric choice, the dedicated-candidate idea), but corrected two issues I found by checking real entity/telemetry data:
1. Its candidate matched `tagKeys: ["k8s.clusterName", "k8s.cluster.name"]` — but a real classic (Infrastructure-agent) k8s-node HOST entity tags the cluster as bare `clusterName`, so that candidate would never have matched a classic node. Fixed here by including all three tag key variants.
2. Its target `extractGuid` used `attribute: entityGuid` — but `kube_pod_info` actually carries the pod GUID as `entity.guid` (dotted). Fixed here.

**Known gap, not addressed in this PR:** in the one OTel/kube-state-metrics cluster I checked, `system.*` metrics have `host.id`/`host.name` null, and no `INFRA-HOST` entity exists for the node at all. Since OTel HOST entities are only synthesized via `host.id` (`entity-types/infra-host/definition.yml`), and kube-state-metrics is a cluster-level scrape that can't know a node's `host.id`, a KSM-only collector setup with no per-node host-metrics receiver may never produce a HOST entity for this rule to link to. That's a separate entity-synthesis gap, not something fixable by adjusting this relationship rule's matching strategy — flagging for visibility rather than silently working around it.

Split out from a combined branch for NR-609078 — see the companion PR for the Node<>Container scoping change (NR-609078 #2): https://github.com/newrelic/entity-definitions/pull/3203

Related ticket: NR-609078

#### Api Review Board (ARB)

ARB Jira ticket:
https://new-relic.atlassian.net/browse/NR-609078

### Checklist

[NR-609078]: https://new-relic.atlassian.net/browse/NR-609078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


entityBot (dismissed):

entityBot (dismissed):

entityBot (approved):

entityBot (dismissed):

entityBot (dismissed):

relationships/synthesis/INFRA-HOST-to-INFRA-KUBERNETES_POD.stg.yml
```diff
@@ -1,59 +1,119 @@
...
+    origins:
+      - OpenTelemetry
+    conditions:
+      - attribute: metricName
+        anyOf: [ "kube_pod_info" ]
+      # regex rather than `present: true`: 3% of kube_pod_info rows carry
+      # k8s.node.name="" (unscheduled pods), which `present` does not screen
+      # and which would look up the empty string.
+      - attribute: k8s.node.name
+        regex: ".+"
```
nr-cmorenoin: This assumption is not true. The `present` clause must work because it check the attribute is not null. Keep in mind that regex is always more expensive

relationships/synthesis/INFRA-HOST-to-INFRA-KUBERNETES_POD.stg.yml
```diff
@@ -1,59 +1,119 @@
...
+            value: INFRA
+          type:
             value: HOST
+            valueInGuid: NA
+          identifier:
+            fragments:
+              - attribute: host.id
+            hashAlgorithm: FARM_HASH
       target:
+        buildGuid:
```
nr-cmorenoin: Both source and target cannot have `buildGuid` at the same time. This relationship is not valid

nr-cmorenoin (commented):
Just a couple of comments to be addressed

entityBot (dismissed):

entityBot (approved):

entityBot (approved):

nr-cmorenoin (approved):

PR #3173, opened 2026-08-17
fix: CONFLUENTCLOUDKAFKATOPIC relationship match fails for private-networked Dedicated clusters

shashank-reddy-nr:
### Relevant information

`apmProducesConfluentCloudKafkaTopicStgOverride`/`apmConsumesConfluentCloudKafkaTopicStgOverride` match `CONFLUENTCLOUDKAFKATOPIC` by comparing `bootstrapServers` against the entity's `confluent.clusterBootstrapServers` tag. On Dedicated clusters with private networking (PrivateLink/PSC/Private Link), the agent-observed broker address never equals that tag — verified against live telemetry on AWS/GCP/Azure — so the match always misses and falls back to `CREATE_UNINSTRUMENTED`.

The `lkc-<id>` cluster ID is still present in the private-networking broker address and matches `confluent.clusterId` exactly, so this adds it as Tier 2 (`bootstrapServers` stays Tier 1, now `NO_OP` on miss instead of short-circuiting) — same tiered-fallback pattern as `AWSDOCDBINSTANCE.yml`.

Changes:
- `APM-APPLICATION-to-KAFKATOPIC.stg.yml` / `relationships/candidates/KAFKATOPIC.stg.yml` — the fix above.
- `confluent_dashboard.json` (production) — unrelated: consumer-lag chart used `latest()`, which renders empty on sparse data; switched to `max()`.

#### Api Review Board (ARB)

ARB Jira ticket: _not yet filed_

### Checklist

* [x] I've read the guidelines and understand the acceptance criteria.
* [x] N/A — no new entity type or identifier introduced.
* [ ] I've linked an ARB ticket & received approval from API Review Board in order to make these changes

PR #3207, opened 2026-08-27
Split Confluent Cloud PrivateLink synthesis rule 

shashank-reddy-nr:
Fixes a regression from #3173. Adding a `clusterId` capture field to the Confluent Cloud synthesis rule broke relationship creation for every public cluster, not just the private ones it was meant to add.

`GuidToLookupMaterializer.extractFields()` aborts the whole rule if any `capture` field fails to regex-match — no optional-field semantics at synthesis time. `clusterId`'s regex only matches PrivateLink hosts (`lkc-*`), so it always failed on public clusters (`pkc-*`), killing the candidate before `candidate-relationship-matcher` ever saw it.

**Synthesis file**: split the combined rule into two — `bootstrapServers`+`topic` (public) and `clusterId`+`topic` (PrivateLink). Their conditions are mutually exclusive (`(?!lkc-)` on the public rule, `lkc-` prefix required on the PrivateLink rule) — verified against a real customer's private cluster, where APM's PrivateLink accesspoint DNS (`lkc-17z27z-ap4ggyrk...`) never matches the entity's real `confluent.clusterBootstrapServers` tag (`lkc-17z27z.eu-west-2.aws.private...`), but `confluent.clusterId` does match on both sides. Without the exclusion, both rules fired for the same private-cluster metric and produced a phantom uninstrumented entity alongside the correct relationship.

**Candidate file**: the bootstrapServers tier's `onMiss` was `NO_OP`, which only made sense when one candidate could fall through to the clusterId tier. Now that the two are disjoint candidates, that fallthrough never happens, so changed it to `CREATE_UNINSTRUMENTED` to match production's original behavior.

Assisted-by: Claude Sonnet 5

### DIFF
--- a/relationships/candidates/KAFKATOPIC.stg.yml
+++ b/relationships/candidates/KAFKATOPIC.stg.yml
@@ -18,7 +18,11 @@ lookups:
       uninstrumented:
         type: KAFKATOPIC
 
-  # Tier 1: bootstrapServers match (public/Basic/Standard clusters).
+  # bootstrapServers match (public/Basic/Standard clusters). Synthesis now emits this as its
+  # own candidate carrying only bootstrapServers+topic (never clusterId), so this tier is
+  # terminal in practice: the clusterId tier below is structurally skipped for it (missing
+  # field), never reached as a fallback. onMiss must create the uninstrumented placeholder
+  # here directly, matching production's unmodified single-tier behavior.
   - entityTypes:
       - INFRA-CONFLUENTCLOUDKAFKATOPIC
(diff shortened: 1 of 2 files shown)


entityBot (dismissed):

entityBot (dismissed):

entityBot (dismissed):

entityBot (approved):

vagrawal-newrelic (approved):
LGTM

bgalmesnr (approved):

PR #3215, opened 2026-09-02
fix: ReplicaSet-to-Pod relationship GUID synthesis

codylawson:
### Relevant information
Source the Kubernetes Integration rule from K8sPodSample + createdKind (matching DaemonSet/StatefulSet/Job) instead of K8sReplicaSetSample, which never matches (real event type is K8sReplicasetSample) and whose entityGuid attribute belongs to the ReplicaSet, not a Pod.

Also include clusterName/namespaceName in the source identifier fragments so the built GUID matches the real ReplicaSet entity's GUID (k8s:<cluster>:<namespace>:replicaset:<name>), and add the missing ":" separators in the OTel rule's identifier fragments to match the entity's own composite identifier convention.

Verified against staging NRDB: real entityKey/entityId for a ReplicaSet only match when cluster+namespace are included in the hash input.

#### Api Review Board (ARB)

ARB Jira ticket:

### Checklist

* [x] I've linked an ARB ticket & received approval from API Review Board in order to make these changes


entityBot (approved):

entityBot (approved):

PR #3197, opened 2026-08-26
fix: promote CONFLUENTCLOUDKAFKATOPIC private-networking fix to production

shashank-reddy-nr:
### Relevant information

Promotes the staging-only fix from #3173 to the production files. That PR fixed `NR-607184`: on Confluent Cloud Dedicated clusters with private networking (PrivateLink/PSC/Private Link), the APM agent-observed `bootstrapServers` never matches the entity's `confluent.clusterBootstrapServers` tag, so the APM → `CONFLUENTCLOUDKAFKATOPIC` relationship always fell back to `CREATE_UNINSTRUMENTED`. #3173 validated a fix in staging; this PR carries the same change into production now that it's been observed working there.

Changes (identical in substance to #3173's `.stg` diff, applied to the base files):

* `relationships/candidates/KAFKATOPIC.yml` — add a Tier 2 (final) candidate lookup for `CONFLUENTCLOUDKAFKATOPIC` matching on `confluent.clusterId` (the `lkc-<id>` prefix, still present on the private-networking broker address). Tier 1 (`bootstrapServers`) changes from `CREATE_UNINSTRUMENTED`-on-miss to `NO_OP`-on-miss so it falls through to Tier 2 instead of short-circuiting. Same tiered-fallback pattern as `AWSDOCDBINSTANCE.yml`.
* `relationships/synthesis/APM-APPLICATION-to-KAFKATOPIC.yml` — `apmProducesConfluentCloudKafkaTopic`/`apmConsumesConfluentCloudKafkaTopic` each gain a `clusterId` fallback field, captured via regex `^(lkc-[a-zA-Z0-9]+)` from the same `metricName__4` attribute `bootstrapServers` already reads.
* `entity-types/infra-confluentcloudkafkatopic/definition.yml` — set `primaryOwner.teamName` to `"OnHost Agent and Integrations"` (was `"no owner"`), matching `infra-kafkacluster`/`infra-kafkabroker`/`infra-kafkatopic`.
* `entity-types/infra-confluentcloudkafkatopic/confluent_dashboard.json` — consumer lag offsets widget: `latest()` → `max()`, so the chart doesn't render empty on sparse data.

`npm run check` in `validator/` passes clean.

#### Api Review Board (ARB)

No new identifiers, entity types, or APIs — same classification as #3173, which entityBot auto-labeled `ARB:not-required`.

ARB Jira ticket: https://new-relic.atlassian.net/browse/NR-611440

### Checklist

[...]

PR #3221, opened 2026-09-03
fix: promote #3207's corrected Confluent Cloud PrivateLink fix to production

shashank-reddy-nr:
## What happened

#3197 was meant to promote the NR-607184 fix (Confluent Cloud PrivateLink clusters not getting APM relationships) to production, but it copied the pre-#3207 diff instead of the corrected one. That version combines `bootstrapServers` and a `clusterId` capture field in a single synthesis rule, which breaks relationship creation for **public** Confluent Cloud clusters too, not just the private ones it was meant to help — `clusterId`'s regex only matches PrivateLink hosts (`lkc-*`), so it fails to capture on public hosts (`pkc-*`) and takes the whole candidate down with it.

This PR applies #3207's already-reviewed fix to the production files instead: split into two independent rules, one for public clusters (`bootstrapServers`+`topic`) and one for PrivateLink clusters (`clusterId`+`topic`), with mutually exclusive conditions so only one ever fires per metric.

## Changes

* `relationships/synthesis/APM-APPLICATION-to-KAFKATOPIC.yml` — split `apmProduces/ConsumesConfluentCloudKafkaTopic` into a public rule and a new `...PrivateLink` rule.
* `relationships/candidates/KAFKATOPIC.yml` — `bootstrapServers` tier's `onMiss` reverts to `CREATE_UNINSTRUMENTED` (matching original production behavior; no more fallthrough needed now that the candidates are disjoint).

## Verification

* `npm run check` passes.
* Regex-tested the new conditions against every broker URL form we could find or reason about — public (AWS/GCP/Azure) and private (direct-networking and PrivateLink/multi-cloud access point, all three clouds) — confirmed public never matches, private always matches exactly one of the Produce/Consume rules.
* Confirmed against a real customer's PrivateLink cluster that `clusterId` matches on both sides even when `bootstrapServers` never will.
* Confirmed on a live production test that public clusters are unaffected by this change (they already worked, and still do).

Same classification as #3173/#3197/#3207 — no new identifiers or entity types.

### DIFF
--- a/relationships/candidates/KAFKATOPIC.yml
+++ b/relationships/candidates/KAFKATOPIC.yml
@@ -18,7 +18,7 @@ lookups:
       uninstrumented:
         type: KAFKATOPIC
   
-  # Tier 1: bootstrapServers match (public/Basic/Standard clusters).
+  # bootstrapServers match (public/Basic/Standard clusters).
   - entityTypes:
     - INFRA-CONFLUENTCLOUDKAFKATOPIC
     tags:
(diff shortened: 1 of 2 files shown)


entityBot (approved):

vagrawal-newrelic (approved):
LGTM

entityBot (approved):

davidgonzalez-nr (approved):

PR #3231, opened 2026-09-09
fix: iIsolate k8s HOST candidate onMiss NO_OP

codylawson:
### Relevant information

Follow-up to #3202 (Host↔Pod GUID resolution, NR-609078). Flagged in Slack: https://newrelic.slack.com/archives/C0BMGJVTJJK/p1788941572652809

`relationships/candidates/HOST.stg.yml` got a 5th lookup rule in #3202 for k8s node matching (used by the `otelKsmK8sHostHostsPod` fallback in `INFRA-HOST-to-INFRA-KUBERNETES_POD.stg.yml`), with `onMiss: NO_OP` — intentional, so a pod that misses the lookup doesn't mint an uninstrumented HOST for every miss across the whole k8s fleet.

The matcher has a known limitation, called out by Milagros Borroni in the thread above: `onMiss: NO_OP` on any single rule in a category suppresses `CREATE_UNINSTRUMENTED` for *every* rule in that category, not just the one that missed. That silently broke uninstrumented `HOST` creation for all the other (non-k8s) HOST candidate rules on staging — APM was one of the teams that noticed the regression.

This PR follows the fix Milagros outlined, based on the existing `CONTAINER_WITHOUT_UNINSTRUMENTED` precedent (`relationships/candidates/CONTAINER_WITHOUT_UNINSTRUMENTED.stg.yml`):

1. Moved the k8s node rule into a new, isolated `relationships/candidates/HOST_WITHOUT_UNINSTRUMENTED.stg.yml` category, keeping `onMiss: NO_OP` scoped to just that rule.
2. Removed the rule from `HOST.stg.yml` — this restores `CREATE_UNINSTRUMENTED` for the other four HOST lookup rules. With the rule gone, `HOST.stg.yml` was byte-identical to the base `HOST.yml`, so the now-redundant staging override file was deleted (staging falls back to the prod file, which is equivalent).
3. Updated `otelKsmK8sHostHostsPod` in `INFRA-HOST-to-INFRA-KUBERNETES_POD.stg.yml` to point its `lookupGuid.candidateCategory` at `HOST_WITHOUT_UNINSTRUMENTED` instead of `HOST`.

No behavior change for the k8s node matching itself — same predicates, same `NO_OP` outcome on miss. The only change is that it no longer leaks into other HOST candidates. Confirmed no other `candidateCategory: HOST` consumer uses the `host.name`/`cluster.name` field pair (the other five all key on `onHostEntityHostname` or `hostId`), so nothing else is affected by the split.

Ran the validator's schema/folder/rule checks and its jest suite locally — all pass.

#### Api Review Board (ARB)

This only touches relationship candidate/synthesis config (no entity type definition changes), tracked under the existing NR-609078 ticket rather than a new ARB ticket:
https://new-relic.atlassian.net/browse/NR-609078

### Checklist

* [x] The value of the attribute marked as `identifier` will be unique and valid. (n/a — no entity type identifier changes)
* [x] I've confirmed that my entity type wasn't already defined. (n/a — no new entity type)


entityBot (approved):

entityBot (approved):

mborroni (approved):

PR #3233, opened 2026-09-10
Updated rules with server address and port for oracle - Staging

sjyothi54:
### Relevant information
Previously, the Oracle receiver passed host.address as a resource attribute (with host.port expected to follow the same convention), and our entity synthesis/relationship rules were built around those names.

In the latest collector build, the receiver now sends server.address and server.port( which follows the otel semantics) by default instead. Updating the definition rules for oracle

This update aligns entity synthesis and the ORACLEDBINSTANCE relationship candidate/rules to the new server.address/server.port attributes.

Fix: this also resolves the multi-instance-per-host issue — previously, with no port tag populated, relationship matching fell back to host-only matching and incorrectly linked a calling app to every Oracle instance on that host. Now that both server.address and server.port are populated and matched together, the relationship correctly resolves to the specific instance being called, even when multiple Oracle instances run on the same host.

#### Api Review Board (ARB)

ARB Jira ticket:

### Checklist

### DIFF
--- a/entity-types/infra-oracledbinstance/definition.stg.yml
+++ b/entity-types/infra-oracledbinstance/definition.stg.yml
@@ -1,0 +1,107 @@
+domain: INFRA
+type: ORACLEDBINSTANCE
+synthesis:
+  rules:
+    - ruleName: infra_oracle_db_instance_entityId
+      identifier: entityId
+      name: entityKey
+      legacyFeatures:
+        useNonStandardAttributes: true
+        overrideGuidType: true
+      prefixedTags:
+        label.:
+          ttl: PT4H
+      conditions:
+        - attribute: eventType
+          value: 'OracleDatabaseSample'
+      tags:
+        integrationName:
+          ttl: P1D
+        integrationVersion:
+          ttl: P1D
+        reportingAgent:
+          ttl: P1D
+        reportingEndpoint:
+          ttl: P1D
+
+
+    - ruleName: infra_oracledbinstance_host_name_nrfork
+      identifier: host.name
+      name: host.name
+      encodeIdentifierInGUID: true
+      conditions:
+        - attribute: eventType
+          value: Metric
+        - attribute: instrumentation.provider
+          value: opentelemetry
+        - attribute: metricName
+          prefix: oracledb.
+      tags:
+        otel.library.name:
+          entityTagName: instrumentation.name
+        otel.library.version:
+          entityTagName: instrumentation.version
+        server.address:
+          entityTagNames: [server.address, endpoint]
+        server.port:
+          entityTagName: server.port
+
+    - ruleName: infra_oracledbinstance_query_sample
+      identifier: host.name
+      name: host.name
+      encodeIdentifierInGUID: true
+      conditions:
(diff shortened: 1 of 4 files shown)


entityBot (approved):

relationships/synthesis/OTEL-APPLICATION-to-INFRA-ORACLEDBINSTANCE.stg.yml
```diff
@@ -0,0 +1,85 @@
...
+    conditions:
+      - attribute: eventType
+        anyOf: [Metric]
+      - attribute: metricName
+        startsWith: newrelicoracledb.
+      - attribute: instrumentation.provider
+        anyOf: [opentelemetry]
+      # This handles both library name formats for backward compatibility
+      - attribute: otel.library.name
+        anyOf: [github.com/open-telemetry/opentelemetry-collector-contrib/receiver/newrelicoraclereceiver]
```
pkulkarni-newrelic: why anyof? it only has one value in the list.

relationships/synthesis/OTEL-APPLICATION-to-INFRA-ORACLEDBINSTANCE.stg.yml
```diff
@@ -0,0 +1,85 @@
...
+    version: "1"
+    origins:
+      - OpenTelemetry
+    conditions:
+      - attribute: eventType
+        anyOf: [Metric]
+      - attribute: metricName
+        startsWith: oracledb.
+      - attribute: instrumentation.provider
+        anyOf: [opentelemetry]
```
pkulkarni-newrelic: same here. anyOf?

entity-types/infra-oracledbinstance/definition.stg.yml
```diff
@@ -0,0 +1,107 @@
...
+          value: opentelemetry
+        - attribute: metricName
+          prefix: oracledb.
+      tags:
+        otel.library.name:
+          entityTagName: instrumentation.name
+        otel.library.version:
+          entityTagName: instrumentation.version
+        server.address:
+          entityTagNames: [server.address, endpoint]
```
pkulkarni-newrelic: are we able to identify the value whether its of server.address or endpoint if both exists or either of them?

pkulkarni-newrelic:
@sjyothi54 As currently in rules, we are using library names as nrfork receiver. will entity be isolated and then discarded after TTL if after a month everything moves to upstream and nrdot points directly to upstream receiver?

tharun0064 (approved):

sjyothi54:
> @sjyothi54 As currently in rules, we are using library names as nrfork receiver. will entity be isolated and then discarded after TTL if after a month everything moves to upstream and nrdot points directly to upstream receiver?

@pkulkarni-newrelic  Yes, we already have the upstream rule, so switching from nrfork to upstream won't discard or recreate the entity, it just keeps refreshing the same one via the upstream rule. The _nrfork rule will simply become unused at that point and can be removed in a follow-up.

davidgonzalez-nr (approved):

PR #3238, opened 2026-09-16
fix(kafka): broker→topic emits MANAGES, drop a dead duplicate rule (stg)

shashank-reddy-nr:
## Summary

nri-kafka's broker→topic rules in this file emitted `SERVES` while every other Kafka provider emits `MANAGES` for the same fact:

| provider | rule | type |
|---|---|---|
| OpenTelemetry | `otelKafkaBrokerManagesOtelKafkaTopicStgOverride` | `MANAGES` |
| AWS MSK | `awsMskBrokerManagesAwsMskTopicForMetricStreams` | `MANAGES` |
| Confluent Cloud | `confluentCloudClusterManagesConfluentCloudKafkaTopic*` | `MANAGES` |
| nri-kafka | `kafkaBrokerServesKafkaTopicStgOverride1/2` | **`SERVES`** |

Measured on staging: **41** `SERVES` edges from nri-kafka brokers against **26** `MANAGES` from OTel and **55** from MSK. Anything traversing broker→topic has to special-case the on-host integration for no reason.

This looks like an oversight rather than a decision — the OTel rule in this same file is already named `...Manages...`, and its production counterpart is still named `otelKafkaBrokerServesOtelKafkaTopic` while also emitting `MANAGES`, so the name was left behind when the type changed.

## Also removes a dead rule

`kafkaBrokerServesKafkaTopicStgOverride2` was identical to rule 1 except that it omitted `valueInGuid: NA`, putting the type into the target GUID. nri-kafka's `KAFKATOPIC` entities carry `NA` in that slot:

- all **23** live nri-kafka topic entities decode with type-slot `NA`
- all **45** live broker→topic edges, across all **12** brokers, resolve to entities with type-slot `NA`

Not one edge was ever produced in the form rule 2 builds, so it only added a second GUID computation per sample. Rule 1 is the one that works and it stays.

## Second commit — container→broker emits HOSTS

`ContainerContainsKafkaBrokerStg` emitted `IS`. Every other on-host integration states container→process as `HOSTS`:

| rule | type |
|---|---|
| `ContainerContainsElasticsearchNode` | `HOSTS` |
| `ContainerContainsMemcachedInstance` | `HOSTS` |
| `ContainerContainsNginxserver` (+Migration) | `HOSTS` |
| `ContainerContainsRabbitmqNode` | `HOSTS` |
| `ContainerContainsRedisinstance` (+Migration) | `HOSTS` |
| `ContainerContainsKafkaBroker` | **`IS`** ← only one |

Confirmed live, not just in YAML: `CONTAINER -HOSTS-> NGINXSERVER` exists in staging.

The split is by origin and is principled. OpenTelemetry states it as `IS` — the broker entity is synthesized from the container's own telemetry, so they are the same thing — and all four OTel rules here keep `IS`, as does `containerIsOtelRedisInstance`. An on-host integration instead discovers a broker process running *inside* a container it did not synthesize: containment, not identity. `IS` also carries entity-aliasing meaning, which isn't what's being asserted.

The rule **name** is unchanged. All seven sibling on-host rules are named `ContainerContains*` while emitting `HOSTS`, so renaming this one would break the convention rather than follow it.

This one is a behaviour change, hence staging only — worth observing here before the same fix goes to production.

## Scope

Staging only. Production is a separate PR so the two can be reviewed and rolled independently.

## Validation

- `npm --prefix validator run check` exits 0
- both files validate against `relationship-synthesis-schema-v1.json`
- rules renamed to match the type they now emit


entityBot (approved):

entityBot (dismissed):

entityBot (approved):

davidgonzalez-nr (approved):

PR #3239, opened 2026-09-16
fix(kafka): drop the `topic` tagKey from the Confluent clusterId candidate tier

shashank-reddy-nr:
## Summary

The Confluent Cloud PrivateLink/PSC candidate tier added in #3221 still resolves to an `UNINSTRUMENTED` placeholder for a customer whose app emits a `clusterId`/`topic` pair that is an **exact match** for the real topic entity's own tags.

Account: Trading Point (2868847). Two topics, 2/2 exact-match failures, so not a naming or regex defect.

## Root cause

Diagnosed by @davidgonzalez against the candidate-tags DB ([thread](https://newrelic.slack.com/archives/CH5MLECNN/p1789549984113969?thread_ts=1789127832.863229&cid=CH5MLECNN)): the generated lookup matches **all predicates *and* all tagKeys within each predicate**. Listing both `confluent.kafka.server.metric.topic` and `topic` therefore demands an entity carrying **both** tags.

- `confluent.kafka.server.metric.topic` — present
- `topic` — never present on any `CONFLUENTCLOUDKAFKATOPIC` entity in the candidate-tags DB

So the tier could not match anything. Dropping `topic` leaves the one key that is actually indexed.

> removing topic from candidates should fix the query. Also seems it won't cause any issue since I see no single `CONFLUENTCLOUDKAFKATOPIC` entity with `topic` tag in our candidate tags DB.

## One thing worth flagging

`docs/relationships/candidate_relationships.md` describes `tagKeys` as an **OR**:

> In this case, entities with the tag keys `aws.bucketName` **or** `aws.s3.BucketName` will be matched

Read against the docs this change is a no-op. The observed query behaviour is AND. Whichever of the two is wrong deserves a follow-up — either the matcher or the docs. This change makes the tier work against the behaviour that is live today.

## Scope

Prod and stg together: both files carry the same tier with the same defect, so splitting them would leave staging unable to reproduce the fix.

The **bootstrapServers tier is deliberately left alone**. It has the identical `tagKeys` pair and is presumably affected the same way, but it is not the tier this customer traverses and it predates #3221. Happy to include it if the platform team confirms the same reasoning applies.

## Note on rollout

Per the thread, affected topic entities also need a reindex before they match — their tags were last indexed before `confluent.clusterId` existed on them, so the DB has no record of it. This definition change is necessary but not sufficient on its own for the existing entities.

## Validation

- `npm --prefix validator run check` exits 0
- both files pass `relationship-candidates-schema-v1.json`


entityBot (approved):

entityBot (approved):

davidgonzalez-nr (approved):

PR #3244, opened 2026-09-18
Add synthesis rule for curated DbServer* events on ORACLEDBINSTANCE (Staging)

tharun0064:
## Summary
- The `nrsp-pipelines-database-components` `ProcessorDbEventNamespace` component renames `db.server.*` log events (`db.server.top_query`, `db.server.query_sample`, `db.server.query_plan`, `db.server.top_procedure`, `db.server.session.wait_sample`) to curated event types (`DbServerTopQuery`, `DbServerQuerySample`, `DbServerQueryPlan`, `DbServerTopProcedure`, `DbServerSessionWaitSample`) and stamps a `namespace` attribute on each event.
- This processor runs generically across DB engines (e.g. Oracle and SQL Server both emit `db.server.*` events), so the resulting curated `eventType` alone does not distinguish the DB engine.
- Adds a new synthesis rule, `infra_oracledbinstance_db_server_events`, to `entity-types/infra-oracledbinstance/definition.stg.yml` that matches any `eventType` with the `DbServer` prefix, gated on `db.system.name: oracle` to scope matching to Oracle events only. Identity/name use `host.name`, consistent with the existing OTel log-based rules in this file, and the new `namespace` attribute is surfaced as an entity tag.

## Test plan
- [x] `npm --prefix validator run validate-schemas` passes with no errors
- [ ] Verify in staging that `DbServerQuerySample`/`DbServerTopQuery`/etc. events with `db.system.name: oracle` synthesize an `ORACLEDBINSTANCE` entity keyed on `host.name`
- [ ] Confirm SQL Server's curated `DbServer*` events (`db.system.name: microsoft.sql_server`) do not match this rule


entityBot (dismissed):

spathlavath (dismissed):

davidgonzalez-nr (dismissed):

entityBot (approved):

sjyothi54 (approved):

davidgonzalez-nr (approved):

